PR #20, opened 2015-01-28
Bugfix: ConnectionImpl::reportError for loops runs out of bounds

SebastianMelinat:
ConnectionImpl::reportError calls ChannelImpl::reportError of every channel. The channel then removes itself from the connection. That caused the for loop to run out of bounds. Fixed by switching to while.


EmielBruijntjes:
Thanks!

PR #21, opened 2015-02-16
Got rid of GCC compiler warnings

SebastianMelinat:
There were a number of compiler warnings due to unused parameters or ignored type qualifiers on returns.
I introduced a file unused.h  and the unused method for unused parameters and removed superflous const for return types.

### DIFF
--- a/amqpcpp.h
+++ b/amqpcpp.h
@@ -37,6 +37,7 @@
 #include <amqpcpp/outbuffer.h>
 #include <amqpcpp/watchable.h>
 #include <amqpcpp/monitor.h>
+#include <amqpcpp/unused.h>
 
 // amqp types
 #include <amqpcpp/field.h>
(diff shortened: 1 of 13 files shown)


EmielBruijntjes:
I do not see these compiler warnings, what flags do you use for compiling? And I don't find thus "unused()" function verry pretty to the eye - is there no other solution to get around this problem?

SebastianMelinat:
We use -Wall -W

I agree that unused isn't pretty. The alternative would be "(void)(parameterName)" or to remove the parameter name in the method definition. That last alternative is, in my opinion, the nicest, but you can't do it like that if the method definition is its declaration.

Sure, in most cases you don't use a parameter on purpose, but it may indicate bugs, so the warning shouldn't be ignored

EmielBruijntjes:
I'm closing this issue, I don't mind that the compiler complains about the unused parameter.

PR #27, opened 2015-04-24
Added Visual Studio 2013 Project and ported to MS C++11.

jhugard:
Minor changes to get working with Visual Studio 2013:
- Moved include files to _include/amqpcpp_ to eliminate need for symlink (not Windows compatible).
- Updated _.gitignore_
- Replaced _endian.h_ with _portable_endian.h_
- Fixed a few compiler warnings.
- Changed **switch(sizeof(T))** to cascading **else if** blocks, because the former won't compile properly on MS C++11.
- **constexpr** is not supported by MS C++11, so converted several uses to inline


include/amqpcpp.h
```diff
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#define NOMINMAX
```
jhugard: Took the easy way out here.
Windows.h normally defines 'min' and 'max' as macros, which prevents those symbols from being used as member functions, etc..  This #define prevents that behavior, but can cause compilation issues in any 3rd party libraries that depend on the legacy behavior (and is hard to track down unless one knows about it).

EmielBruijntjes:
Based on your input, I've made a couple of changes to the library. Not every change has been merged, because I don't know if I want to include all these Visual Studio files in the project, because I cannot maintain such files, but most changes to the source code have been merged. Can you check this?

jhugard:
Will take a look when I have some time.  Pretty sure there was another constexpr and another use of min/max…






Sent from Windows Mail





From: Emiel Bruijntjes
Sent: ‎Friday‎, ‎April‎ ‎24‎, ‎2015 ‎1‎:‎48‎ ‎AM
To: CopernicaMarketingSoftware/AMQP-CPP
Cc: James Hugard





Based on your input, I've made a couple of changes to the library. Not every change has been merged, because I don't know if I want to include all these Visual Studio files in the project, because I cannot maintain such files, but most changes to the source code have been merged. Can you check this?

PR #29, opened 2015-05-20
Rabbitmq tutorial

RPG-18:
Hi, maybe it will be useful.


EmielBruijntjes:
Thanks for your examples, I've not tested it, but it might be useful for others. I've moved it to the examples/ directory.

Issue #41, opened 2015-09-16
Library


kargapoltsev:
Hi. How to compile a static or dynamic library from AMQP-CPP? I tried to do it using the MVS 2015 but I received a huge number of obscure bugs ...

Kojoley:
I had luck to compile AMQP-CPP with MinGW and MSVC14 (VS 2015) after some changes. Here is patch https://gist.github.com/Kojoley/e5e6b56a375f22ec8ce0

EmielBruijntjes:
I would be nice to receive a pull request to fix this for Windows. I only use, test and have access to Linux environments.

PR #45, opened 2015-09-29
need add compiler options

tangkingchun:
add_definitions( " -std=c++11" )

### DIFF
--- a/examples/rabbitmq_tutorials/CMakeLists.txt
+++ b/examples/rabbitmq_tutorials/CMakeLists.txt
@@ -2,6 +2,7 @@ set(Boost_USE_STATIC_LIBS ON)
 set(Boost_USE_MULTITHREADED ON)  
 set(Boost_USE_STATIC_RUNTIME OFF) 
 find_package(Boost 1.55 REQUIRED COMPONENTS system)
+add_definitions( " -std=c++11" )
 
 if(NOT Boost_FOUND)
     message( FATAL_ERROR "boost must be installed")


Kojoley:
This is completely wrong way to do this, because it's not compiler aware. And this is already done. AMQP-CPP for this purpose uses `set_cxx_norm.cmake`. Pay attention to the following lines in the root `CMakeLists.txt`
```CMake
# ensure c++11 on all compilers
include(set_cxx_norm.cmake)
set_cxx_norm (${CXX_NORM_CXX11})
```

Btw with CMake 3.1 it can be achieved as follow:
```CMake
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

EmielBruijntjes:
I dont know how this CMakefile stuff works, because I always use the regular Makefile. Should I merge this, or forget about it?

Kojoley:
I'll suggest not to merge this but replace `set_cxx_norm.cmake` with something like this

```CMake
# enable c++11
if(MSVC)
  if(MSVC_VERSION LESS 1800)
    # https://msdn.microsoft.com/en-US/library/hh567368.aspx
    message(FATAL_ERROR "Visual Studio 2015 is required for constexpr")
  else()
    # do nothing as Visual Studio have enabled C++11 by default
  endif()
elseif(CMAKE_VERSION VERSION_LESS "3.1")
  if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" OR
      "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
    if (NOT ("${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+0x" OR
             "${CMAKE_CXX_FLAGS}" MATCHES "-std=gnu\\+\\+0x" OR
             "${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+11" OR
             "${CMAKE_CXX_FLAGS}" MATCHES "-std=gnu\\+\\+11"))
      if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
        if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.7")
          set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
        else()
          set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
        endif()
      else()
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
      endif()
    endif()
  endif()
else()
  set(CMAKE_CXX_STANDARD 11)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
endif()
```

soloveyhappy:
First of all thx for a great job. I have one questionIs it posible to build project using MinGW? I had some problems with posix sockets and netdb lib.

Kojoley:
@soloveyhappy yes, but with some changes (details here #41) and not a mainstream version (because of `TcpConnection` which uses POSIX sockets as you already noticed).

surgura:
I think we fixed this as of 91c4f891590c630df4f2a18e3f5925ddf41b3d3f. We now require CMake `VERSION 3.1` so we can use `set (CMAKE_CXX_STANDARD 11)`.

PR #46, opened 2015-10-02
Fix Windows build

jstefanski:
Fixed windef.h shadowing of 'max' definition and added LEAN_AND_MEAN
definition for more lightweight build.
Added missing 'functional' include in includes.h file.

This pull request fixes errors:
```
Error	C2039	'function': is not a member of 'std'	amqp-cpp	amqp-cpp\include\callbacks.h	22
Error	C2039	'function': is not a member of 'std'	amqp-cpp	amqp-cpp\include\callbacks.h	22
```

```
Error	C2039	'function': is not a member of 'std'	amqp-cpp	amqp-cpp\include\callbacks.h	22
```


EmielBruijntjes:
I've merged your code, but I'm sure a lot of stuff has been broken in the meantime because I've added a TCP module to the AMQP-CPP library that involves linux-specific system calls. Would be nice if you could help me out here and fix it for windows.

PR #47, opened 2015-10-21
The server heartbeat is stored in Connection

SebastianMelinat:
Under some circumstances it can happen that our AMQP client does not receive any more messages, including heartbeats, without being able to detect an AMQP- or network error. To fix this our client expects heartbeats and resets the connection after x heartbeats were missed. To know when to expect a heartbeat I added the possibility to retrieve the servers heartbeat duration by storing it in the Connection object.

### DIFF
--- a/include/connection.h
+++ b/include/connection.h
@@ -154,6 +154,15 @@ public:
     }
 
     /**
+    *  Retrieve the heartbeat delay used by this connection
+    *  @return uint16_t
+    */
+    uint16_t heartbeat() const
+    {
+        return _implementation.heartbeat();
+    }
+
+    /**
      *  Some classes have access to private properties
      */
     friend class ChannelImpl;
--- a/include/connectionimpl.h
+++ b/include/connectionimpl.h
@@ -94,6 +94,12 @@ protected:
     std::queue<OutBuffer> _queue;
 
     /**
+    *  Heartbeat delay
+    *  @var uint16_t
+    */
+    uint16_t _heartbeat = 0;
+
+    /**
      *  Helper method to send the close frame
      *  Return value tells if the connection is still valid
      *  @return bool
@@ -340,6 +346,23 @@ public:
     std::size_t channels() const
     {
         return _channels.size();
+    }
+
+    /**
+     *  Heartbeat delay
+     *  @return uint16_t
+    */
+    uint16_t heartbeat() const
+    {
+        return _heartbeat;
+    }
+
+    /**
+     *  Set the heartbeat delay
+    */
+    void setHeartbeat(uint16_t heartbeat)
+    {
+        _heartbeat = heartbeat;
     }
 
     /**
--- a/src/connectiontuneframe.h
+++ b/src/connectiontuneframe.h
@@ -137,6 +137,9 @@ public:
         // theoretically it is possible that the connection object gets destructed between sending the messages
         Monitor monitor(connection);
         
+        // store the heartbeat the server wants 
+        connection->setHeartbeat(heartbeat());
+
         // send it back
         connection->send(ConnectionTuneOKFrame(channelMax(), frameMax(), heartbeat()));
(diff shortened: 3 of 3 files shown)


EmielBruijntjes:
I have merged your pull request - but I think you better take a good look at your code, as the problem is probably there. I have never experienced any problems with heartbeats, and the connections here are kept open foro months without problems. Are you sure that you use a non-blocking event loop, and that you application does not block on other operations (like reading a very big file)? If your application hangs for a while and "forgets" to process incoming data from the RabbitMQ server, or is too late in processing it, you do not respond in time on the heartbeat frames.

SebastianMelinat:
The problem doesn't seem to be our client. The client remains responsive, as demonstrated by the simple fact that it's able to detect the missing heartbeats. The problem is that the broker just stops sending data. We haven't figured out the reason for that yet, and as the problem only occurs seldomly and, seemingly, without reason, using the heartbeat seemed to be the best solution. After all, RabbitMq suggests to use the heartbeat like that.

SebastianMelinat:
The problem isn't our client, that remains responsive, as demonstrated by the simple fact that it's able to detect the missing heartbeats. The problem is a load balancer that we use in the system. We haven't figured out the exact reason for the problem yet, and as the problem only occurs seldomly and, seemingly, without reason, using the heartbeat seemed to be the best solution. And after all, RabbitMq suggests to use the heartbeat like that.

Kojoley:
This can happen because of broken connection in case if you only consume queues so pending network read will not detect broken connection and will wait forever.
(As you don't initiate any writing which will loose ACKs on dead connection and will lead to closed socket in short time by OS)

EmielBruijntjes:
If you use the AMQP-CPP library correctly, there is no pending network read. In your application main event loop you monitor whether the AMQP filedescriptor is readable, and once it is, you read the data from it (in a non-blocking way), and pass it on to the AMQP-CPP library. The AMQP-CPP detects whether the incoming data contains a heartbeat frame, and if it does, creates a response-frame and sends it back to your onData() method. You are then supposed to send that frame over a socket connection back to the RabbitMQ server.

Even when you're just consuming, you should still run this event loop, and read in all data that comes in via the socket and pass it on to the AMQP-CPP library. No way that you can miss a heartbeat or become idle for too long.

If you use the new AMQP::TcpConnection and AMQP::TcpChannel classes instead, it is even easier, because the socket-reading and socket-writing no longer has to be implemented by the library-user, but you only have to include the filedescriptor in your main event loop, and call TcpConnection::process() every time that you see that it is readable and/or writable.

SebastianMelinat:
That's exactly what I do, only using the Qt Library.

Should we find the exact problem and a solution to why our broker stops broadcasting I'll post a solution here. Although I think it's a very specific problem that isn't relevant to most.

Kojoley:
You will miss a heartbeat if connection is zombie in any way no matter how you work with network. OS can't detect broken connection until any write operation is failed (due to missed ACK for it). So you have two ways:
* Enable TCP keep-alive
* Monitor heartbeat with timer (or reset timer on every received heartbeat)

SebastianMelinat:
That's true. If the connection is dead but not broken on the server side the client won't receive an AMQP error and won't detect a network error, as its connection is still alive.
That was our problem. The service the AMQP client was embedded in was still doing it's job and responding to API calls without problem, it just didn't receive anything from the AMQP server.
I'm not sure if Qt has TCP keep-alive, but anyway I opted for the second possibility @Kojoley mentions. 

See also here: https://www.rabbitmq.com/heartbeats.html
"Heartbeat frames are sent about every timeout / 2 seconds. After two missed heartbeats, the peer is considered to be unreachable. Different clients manifest this differently but the TCP connection will be closed. **When a client detects that RabbitMQ node is unreachable due to a heartbeat, it needs to re-connect.** "

PR #51, opened 2015-12-04
Fix compiling on OS X Yosemite

indrekj:
Before this
```
...
./pipe.h:45:13: error: use of undeclared identifier 'pipe2'
        if (pipe2(_fds, O_CLOEXEC) == 0) return;
            ^
In file included from tcpconnection.cpp:14:
In file included from ./tcpresolver.h:22:
In file included from ./tcpconnected.h:19:
./tcpbuffer.h:359:27: error: use of undeclared identifier 'writev'; did you mean 'write'?
            auto result = writev(socket, (const struct iovec *)&buffer, index);
                          ^~~~~~
                          write
/usr/include/unistd.h:490:10: note: 'write' declared here
ssize_t  write(int, const void *, size_t) __DARWIN_ALIAS_C(write);
         ^
2 warnings and 2 errors generated.
make[1]: *** [tcpconnection.o] Error 1
make: *** [all] Error 2
```

**NB**: I haven't tested it properly. I tried running tests but couldn't because of the closed source library.

### DIFF
--- a/src/pipe.h
+++ b/src/pipe.h
@@ -42,8 +42,16 @@ public:
     Pipe()
     {
         // construct the pipe
+#ifdef _GNU_SOURCE
         if (pipe2(_fds, O_CLOEXEC) == 0) return;
-        
+#else
+        if (
+            pipe(_fds) == 0 &&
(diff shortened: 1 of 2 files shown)


elthariel:
:+1:

PR #54, opened 2016-01-09
Add TcpHandler implementation for libevent

brentnd:
I implemented LibEventHandler following the pattern of the LibEvHandler which plugs TcpHandler into the libevent library. I tested it manually with tests/libevent.cpp with create queues, consume, etc. Also verified no memory leaks with valgrind.

Requires installation of libevent.


EmielBruijntjes:
Thanks for doing this, it's a very welcome contribution.

PR #62, opened 2016-04-02
Fix #52

### LINKED ISSUE #52: Connection failed on armv7l target
I'm trying to use amqp-cpp on armv7l (gcc 4.9.3 / uclibc-1.0.6). On host machine x86_64 program works as excepted, but on arm I get this error:

```AMQP error invalid end of frame marker```

Is there any experience with amqp-cpp on non-x86 platforms?

upd: I found that all the negative values in the buffer are wrong. For example, 206 instead of -50.
The problem is that the char may be signed or unsigned depending on the compiler

oschmi:
As mentioned by half45 the solution was to replace char by int8_t for armv7.
I was able to test this on raspberries with gcc 5.2.1. For me it works fine.
Would love if you can confirm this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,10 +30,8 @@ install(TARGETS amqp-cpp
         ARCHIVE DESTINATION lib
 )
 
-option(BUILD_TUTORIALS "build rabbitmq tutorials" OFF)
-if(BUILD_TUTORIALS)
-    add_subdirectory(examples/rabbitmq_tutorials)
-endif()
+option(BUILD_TUTORIALS "build rabbitmq tutorials" ON)
+add_subdirectory(examples/rabbitmq_tutorials)
 
 set(AMQP-CPP_INCLUDE_PATH ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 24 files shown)


EmielBruijntjes:
Hey, thanks for this pull request. But I am not convinced that this is the right approach.

Indeed, there are a number of places in the AMQP-CPP library where we use "char" for storing numbers. And this indeed causes problems on some systems that treat this as signed or unsigned data, while one other systems it is handled differently. It would indeed be an improvement to change this to "int8_t" or "uint8_t" to fix this.

However, in many other places "char" is being used for storing characters - and I think that "char" is a very appropriate data type for characters. Is it possible to change your pull request in such a manner that we only use "(u)int8_t" in case we use numeric values that happen to be stored in 8 bits, and "char" for character data?

oschmi:
I dont really looked that deep into AMQP-CPP to see where you use char for storing numbers. However you could maybe change only the signatures where numbers are used and cast it to "char" or "int8_t" while calling.

Another approach is maybe to use "signed char" explicitly? I think "signed char" is also a appropriate data type ;)

martijnotto:
We don't need to change all appearances of char to uint8_t. In most places it seems to make sense to use char. I changed it in the place where it matters which solves the bug.

PR #72, opened 2016-06-25
Some portability fixes for Solaris-like systems

XeCycle:
Hi, two commits for Solaris portability.

1. `FIONREAD` is defined in `sys/filio.h`
2. the standard C++ library is missing a template overload for `pow`

The fix for 2 is a bit ugly, though.

### DIFF
--- a/include/decimalfield.h
+++ b/include/decimalfield.h
@@ -108,7 +108,7 @@ public:
     virtual void output(std::ostream &stream) const override
     {
         // output floating point value
-        stream << "decimal(" << _number / pow(10, _places) << ")";
+        stream << "decimal(" << _number / pow(10.0f, _places) << ")";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


include/decimalfield.h
```diff
@@ -149,7 +149,7 @@ class DecimalField : public Field
      */
     virtual operator float() const override
     {
-        return static_cast<float>(_number / pow(10, _places));
+        return static_cast<float>(_number / pow(10., _places));
```
Kojoley: I think it's better to avoid casting vetween float and double and just use float literal like `10.0f` or directly select specialized version of pow like `pow<float>()` here

XeCycle:
@Kojoley changed to `10.0f`.

Btw also changed author email to my github one.

martijnotto:
Could pow() be called with a long as well? If possible, I'd find that a much cleaner solution.

XeCycle:
Tried 10ul but it does not work, sorry.  Seems the std expects `base` to be some floating-point type anyway.

martijnotto:
It seems solaris is technically within it's right. Most likely glibc is overly nice here by providing extra overloads not required by the standard, so technically the old code was wrong and your PR fixes it.

PR #73, opened 2016-06-26
Bugfix: add missing include "includes.h" for deferredconsumerbase.cpp

javeme:
When I compiled on the windows, there was a lot of errors like:
  error C2065: "xxx": undeclared identifier...

### DIFF
--- a/src/deferredconsumerbase.cpp
+++ b/src/deferredconsumerbase.cpp
@@ -10,6 +10,7 @@
 /**
  *  Dependencies
  */
+#include "includes.h"
 #include "../include/deferredconsumerbase.h"
 #include "basicdeliverframe.h"
 #include "basicheaderframe.h"


martijnotto:
I don't really see what messages you could have gotten from this - all dependencies are included right from deferredconsumerbase.cpp - but it can't really hurt to add the includes.h (it only makes compilation slightly slower), so I have merged it.

javeme:
Sorry I should paste some error messages:
include\deferred.h (97): error C2061: syntax error: identifier "uint32_t"
include\outbuffer.h (185): error C3861: "htobe64": identifier not found
...


uint32_t from stdint.h (include in https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/master/src/includes.h#L13)
htobe64 from endian.h (include in https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/master/src/includes.h#L35)

martijnotto:
Could you make a pull request that adds the relevant includes to the correct files?

PR #76, opened 2016-07-01
a couple of fixes

maxim-ky:
Hi guys,

At first, thank you for such a great work. AMQP-CPP is nice in terms of architecture. it's also well-documented. And your work is especially valuable because AMQP-CPP is the only asynchronous AMQP c++ library.

But it seems you have not enough resources to support multiple platforms/compilers because I faced a lot of issues during compiling this one under GCC 4.7.3/cmake 2.8. If this toolset is considered to be outdated, then just don't merge 40ebb42.
Other commits are more important, mainly the undefined-behavior one.

Regards

### DIFF
--- a/src/connectionimpl.cpp
+++ b/src/connectionimpl.cpp
@@ -123,7 +123,8 @@ uint64_t ConnectionImpl::parse(const Buffer &buffer)
         try
         {
             // try to recognize the frame
-            ReceivedFrame receivedFrame(ReducedBuffer(buffer, processed), _maxFrame);
+            ReducedBuffer reduced_buf(buffer, processed);
+            ReceivedFrame receivedFrame(reduced_buf, _maxFrame);
             
             // do we have the full frame?
             if (receivedFrame.complete())
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ add_subdirectory(src)
 add_subdirectory(include)
 
 add_library(amqp-cpp STATIC ${SRCS})
-target_include_directories(amqp-cpp SYSTEM PUBLIC ${PROJECT_SOURCE_DIR})
+include_directories(${PROJECT_SOURCE_DIR})
 install(TARGETS amqp-cpp
         ARCHIVE DESTINATION lib
 )
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PREFIX                  = /usr
+PREFIX                  ?= /usr
 INCLUDE_DIR             = ${PREFIX}/include
 LIBRARY_DIR             = ${PREFIX}/lib
 export LIBRARY_NAME		= amqpcpp
--- a/include/buffer.h
+++ b/include/buffer.h
@@ -32,6 +32,11 @@ class Buffer
 {
 public:
     /**
+    *  Destructor
+    */
+    virtual ~Buffer() {}
+    
+    /**
      *  Total size of the buffer
      *  @return size_t
      */
--- a/include/connectionhandler.h
+++ b/include/connectionhandler.h
@@ -37,6 +37,11 @@ class ConnectionHandler
 {
 public:
     /**
+    *  Destructor
+    */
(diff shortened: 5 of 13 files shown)


src/connectionimpl.cpp
```diff
@@ -123,7 +123,9 @@ uint64_t ConnectionImpl::parse(const Buffer &buffer)
         try
         {
             // try to recognize the frame
-            ReceivedFrame receivedFrame(ReducedBuffer(buffer, processed), _maxFrame);
```
Kojoley: Why do you think that It is an undefined behaviour?

12.2/5 "Temporary objects":

> The temporary to which the reference is bound or the temporary that is the complete object to a subobject of which the temporary is bound persists for the lifetime of the reference except as specified below. A temporary bound to a reference member in a constructor’s ctor-initializer (12.6.2) persists until the constructor exits. **A temporary bound to a reference parameter in a function call (5.2.2) persists until the completion of the full expression containing the call.**

maxim-ky: > Why do you think that It is an undefined behaviour?

The next statement is:
`
if (receivedFrame.complete())
`
ReducedBuffer is destroyed here because that **full expression containing the call** is already evaluated.
But ReceivedFrame::complete uses stored reference to the destroyed ReducedBuffer.
`
bool ReceivedFrame::complete() const
{
    return _buffer.size() >= _payloadSize + 8;
}
`

martijnotto:
Thanks for the pull-request. I was wondering if there is a specific reason that you added all the virtual destructors to all the interface classes. As far as I can see we never destruct them in the library but I maye not have looked properly. Did you find instances where this is actually done?

If they really are necessary could you please add the docblocks to them. Just a simple

    /**
     *  Destructor
     */

will do.

src/connectionimpl.cpp
Kojoley: The _full expression containing the call_ is [here](https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/f3343ad1cb64f5a58849fcf0ee3cc22dece2393c/src/connectionimpl.cpp#L126-150).
Look at example in the [C++ Standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf) on page 254.

maxim-ky:
> I was wondering if there is a specific reason that you added all the virtual destructors to all the interface classes.

The reason is this GCC warning: *a class has virtual functions but non-virtual destructor.*

> As far as I can see we never destruct them in the library but I maye not have looked properly.

You're right. Mostly it's the matter of some common code style. 
a Herb Sutter's quote:
> Guideline #4: A base class destructor should be either public and virtual, or protected and nonvirtual.

the original link: http://www.gotw.ca/publications/mill18.htm

Does that sound appropriate to you?

src/connectionimpl.cpp
maxim-ky: I'm sorry, but that example looks irrelevant to me (there's nothing about saving a reference).
C++ expressions can't consist of several statements (not willing to quote Standard for you, sorry again)

Actually, you could compile that code snippet with an existing compiler and see the output.
http://pastebin.com/UcLrCkvL

Kojoley: You are making copy of object here `B bbb = B( A() );`. It is completely different situation and does not belong to what you have in your patch.

maxim-ky: Actually, copying doesn't occur here. I used that syntax to avoid so called *most vexing parse*, but if you have some doubts, you could use this code snippet which completely resembles the original code.
http://pastebin.com/erL8GGXs

martijnotto:
That sounds perfectly logical to me. Could you update your code to include a simple docblock for the destructors? I'd be happy to merge it then.

maxim-ky:
> Could you update your code to include a simple docblock for the destructors?

It's done.

martijnotto:
Looking good. Thanks for the help.

PR #77, opened 2016-07-03
Correct the returning type of Connection::waiting()

javeme:
The returning type of Connection::waiting() should be bool,
but currently it's size_t type.


martijnotto:
This makes a lot more sense. Thanks.

PR #80, opened 2016-07-10
Avoid SIGPIPE on OS X without MSG_NOSIGNAL

theirix:
OS X does not have Linux-specific option MSG_NOSIGNAL.
SIGPIPE can be avoided on OS X by using sockopt SO_NOSIGPIPE.
Credits: https://noahdesu.github.io/2014/01/16/port-sendmsg.html

Project compiles on both OS X and Linux. Had not tested yet but it seems to be ok.

### DIFF
--- a/include/tcpdefines.h
+++ b/include/tcpdefines.h
@@ -1,0 +1,27 @@
+#pragma once
+
+/**
+ * No MSG_NOSIGNAL on OS X.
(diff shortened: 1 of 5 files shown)


theirix:
Hello!
Any news about failing OS X builds and this fix?

EmielBruijntjes:
Thanks for contributing, this one missed my attention.

PR #85, opened 2016-07-16
Implement copy constructor for class Monitor

javeme:
To support the instance copy of Monitor like the following:
```
    AMQP::Monitor monitor(consumer);
    consumer->bindExchange(exchange, AMQP::direct).onSuccess([&, monitor](){
        // do something when the consumer is ready!
        if(monitor.valid())
            consumer->doSomething();
    });
```

### DIFF
--- a/include/monitor.h
+++ b/include/monitor.h
@@ -58,6 +58,16 @@ public:
     }
 
     /**
+     *  Copy constructor
+     *  @param  monitor
+     */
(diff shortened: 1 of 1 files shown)


EmielBruijntjes:
The code looks fine. But be aware: the AMQP::Monitor class was intended for internal use inside the AMQP-CPP library. In your "user space" code you normally do not need it.

PR #86, opened 2016-07-17
libuv drop in implementation for TcpHandler

ogapo:
Got this working in a project I'm working on figured I'd pay it forward.

Small notes: 
- I had to ifdef out the cast to Array in Array and cast to Table in Table (both have a note about them not being needed by the compiler). Clang actively warns on them and I compile with warnings as errors so that was mildly annoying (would like to use stock). What's the rationale for these?
- Including the external dependency (#include <uv.h>) in this file is actually something I had to remove locally because I don't include it from the system path. I've included it here for consistency with libev.h and libevent.h but I'd like to suggest they (all 3) be removed from the .h for flexibility. if someone is using libX.h and hasn't already included that it will fail to compile in an obvious way.

### DIFF
--- a/include/libuv.h
+++ b/include/libuv.h
@@ -63,7 +63,7 @@ private:
 
             // tell the connection that its filedescriptor is active
             int fd = -1;
-            uv_fileno(static_cast<uv_handle_t*>(handle), &fd);
+            uv_fileno(reinterpret_cast<uv_handle_t*>(handle), &fd);
             connection->process(fd, uv_to_amqp_events(events));
         }
 
@@ -107,9 +107,9 @@ private:
             uv_poll_stop(_poll);
 
             // close the handle
-            uv_close(static_cast<uv_handle_t*>(_poll), [](uv_handle_t* handle) {
+            uv_close(reinterpret_cast<uv_handle_t*>(_poll), [](uv_handle_t* handle) {
                 // delete memory once closed
-                delete static_cast<uv_poll_t*>(handle);
+                delete reinterpret_cast<uv_poll_t*>(handle);
             });
         }
 
--- a/include/libuv.h
+++ b/include/libuv.h
@@ -1,0 +1,220 @@
+/**
+ *  LibUV.h
+ *
+ *  Implementation for the AMQP::TcpHandler that is optimized for libuv. You can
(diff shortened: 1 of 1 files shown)


martijnotto:
Thank you for taking the time to write this and sharing it. A couple of notes though:

- The warning about the unused casting operator is - I believe - incorrect. The whole point is that the casting operators are virtual. You might get an AMQP::Array back as a Field* or Field& somewhere and if the override is not installed the casting operator is simply going to return an empty AMQP::Array (see src/field.cpp). If I remember correctly I already created a bug-report with clang for this.

- The _poll member you have in the AMQP::LibUvHandler class, why is it a pointer? Wouldn't it be better to just use a stack member here? Can you fix this (or tell me the reason it has to be a pointer)? If it absolutely has to be a pointer, why not an std::unique_ptr?

About the casting operators, this could probably be easily solved by making those two problematic casting operators non-virtual and just do a dynamic_cast to see whether it actually is the type we're casting to and then returning it. This would work around the bug you experience in clang.

ogapo:
No problem, happy to help. I do want to mention this lib is great. I spent 3 days wrestling with the (official?) rabbitmq-c library trying to get it to do what I want using libuv worker threads and finally gave up. Found AMQPCPP and had my project up and running with it in hours. Kudos.

Re: your questions
- Good point, I did not notice the virtual there. That makes a lot more sense and I'd better undo my ifdef :-). I have since switched to including amqpcpp using <> so I think that should make it not treat those warnings as errors for me anymore. I'd rather do that than enable RTTI. Thanks for the info
- This is unfortunately necessary. As a rule I don't use direct memory allocation in my project either, but for this case I make an exception. libuv utilizes hidden members in the struct that you allocate (bit of a leaky abstraction imho) and it's not guaranteed to be done with them until the callback to uv_close has fired. Since we're closing the poll in the dtor, there's not really a good way to make the operation async. To handle this we just orphan the pointer and then set a lambda to do the cleanup when libuv is done too. It is guaranteed that no more callbacks fire after uv_close has been called (not necessarily completes) so this should be safe. This pattern unfortunately precludes the use of unique_ptr too since the lambdas are really just function pointers in this case (cannot have captures).

martijnotto:
Ah, I didn't know that callback could be asynchronous itself, that really sucks. I've merged the PR. Thanks again.

PR #93, opened 2016-09-01
Add SASL EXTERNAL authentication mechanism

AnMakc:
Added support for external (e.g. X509 client cert) authentication mechanism.
Code is backward-compatible (old usages of Login will keep working).

To authenticate using SASL EXTERNAL mechanism, pass following object as Login to Connection constructor:
`AMQP::Login(AMQP::LOGIN_EXTERNAL)`

### DIFF
--- a/include/login.h
+++ b/include/login.h
@@ -22,15 +22,6 @@
 namespace AMQP {
 
 /**
- *  Login mechanism enum
- */
-typedef enum
-{
-    LOGIN_PLAIN,
-    LOGIN_EXTERNAL
-} LoginMechanism;
-
-/**
  *  Class definition
  */
 class Login
@@ -57,6 +48,15 @@ private:
 
 public:
     /**
+     *  Login mechanism enum
+     */
+    typedef enum
+    {
+        LOGIN_PLAIN,
+        LOGIN_EXTERNAL
+    } LoginMechanism;
+
+    /**
      *  Default constructor
      */
     Login() : _mechanism(LOGIN_PLAIN), _user("guest"), _password("guest") {}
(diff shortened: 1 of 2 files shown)


include/login.h
```diff
@@ -44,15 +59,20 @@ class Login
     /**
      *  Default constructor
      */
-    Login() : _user("guest"), _password("guest") {}
+    Login() : _mechanism(LOGIN_PLAIN), _user("guest"), _password("guest") {}
```
Kojoley: You can use delegating constructor instead of copy-pasting

include/login.h
```diff
@@ -22,12 +22,27 @@
 namespace AMQP {
 
 /**
+ *  Login mechanism enum
+ */
+typedef enum
```
Kojoley: You can move the declaration to the class

AnMakc:
Code refactored.

trevorperrin:
What is the status of this Pull Request?  This is something I was going to add, but found there was already a request for it.

EmielBruijntjes:
The status is that I dislike the code, but I have not yet had time to dig into this. I dont know how this sasl is supposed to work. But I suspect that we need some sort of base class that is extended by a plain login system and by a sasl login system.

troian:
@EmielBruijntjes this PR work with certificate authentication

henris42:
Hi, this PR is also needed along with 224 for certificate auth to work. There is a bit missing still though - method to easily define when we use EXTERNAL mechanism. This is currently hidden in AMQP::Address. My current address.h test code assumes external auth at the moment when there is no username/pw in the URL. I'm sure there is nicer way to do that, but that would need bit more refactoring.

EmielBruijntjes:
I still don't like this code. The Login class is here to contain a username + password. If you don't need this, it makes more sense to not even construct it, doesn't it?

henris42:
Neither do I. but we do need to give the login type PLAIN vs EXTERNAL when establsihing session. That is ugliness in the protocol itself. Its a stupid thing that makes no sense - but we still have it!
I'm all for to abstract this nicely.

stertingen:
Are there currently ongoing developments on this topic? I'm willing to submit a PR introducing a `Authentication` base class with derived classes `Login` and `External` which provide the `mechanism` and the `response` field for the `ConnectionStartOkFrame`.

EmielBruijntjes:
That makes sense, I will check it when you send in that request

plopp:
@AnMakc I think https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/390 replaces this PR so it can be closed, agree?

AnMakc:
> @AnMakc I think #390 replaces this PR so it can be closed, agree?

Sure, sounds reasonable.

PR #94, opened 2016-09-02
Fixed warnings: self-casts, const qual

theirix:
These patches fixed some annoying warnings in header files.
Could you please check them with your tests and merge it?

### DIFF
--- a/include/array.h
+++ b/include/array.h
@@ -218,16 +218,6 @@ public:
         // postfix
         stream << ")";
     }
-
-    /**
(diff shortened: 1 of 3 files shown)


martijnotto:
I have just put the casting operators back. Removing these is not safe as they are virtual and override the casting operator in the base class (Field).

theirix:
You are right, commit "Removed unneeded self-cast operators" is wrong. Sorry for the misleading patch.

Issue #48, opened 2015-11-06
Build on Windows


Kojoley:
There are some minor and major issues with building on Windows:
1. #46 fixed errors, but somewhere are still defined `max` macros which produces warnings.
2. Missing `htonll` and `ntohll` (mentioned in #41) in `Winsock2.h`, but [MSDN](https://msdn.microsoft.com/en-US/library/windows/desktop/jj710199%28v=vs.85%29.aspx) claims them should be there.
3. `TcpConnection` uses POSIX sockets (mentioned in #46 and #45).
4. Warnings in Visual Studio (because of more verbose warning level than GCC has by default)

EmielBruijntjes:
I am not a Windows user. Can you send a pull request that fixes this?

gk0090:
I am also getting that kinda error and i replace socket.h with winsock2.h and netdb.h,unistd.h with windows.h. now i am getting more then thousands error.Most of them are related to different callback class.

Please help me with this.
Thanks

dharanikumar-s:
Hi, I too get the same error as @gk0090 commented. I too tried replacing linux header files with windows equivalents ones, but of no use. Is there any solution available for this issue? Do you require the detailed build log for debugging this issue?

Vermillionaire:
I feel like this section in the README is a bit ambiguous

> For users on a non-Linux environment: this library is known to work on your environment too (after all, it does not do any operating specific system calls), but it might take some extra effort to get your compiler to compile it. Please send in your pull requests once you have it running, so that others can benefit from your experiences.

It actually does have some Linux/Unix specific system calls in the pipe.h header. This will probably work fine on a Mac, but Windows pipes are different enough that this could be quite a headache to fix. I'd probably rephrase this saying that it should support most Linux/Unix systems, but currently not Windows. Just a friendly suggestion :smile:

zouqiang8233:
Remove the Linux platform for the pipe use simultaneous TCP connection will be some differences between the socket under the relevant functions to modify I passed the GIT to submit these But why I can't show

dqueffeulouatw:
Finally, is someone can confirm this lib works on Windows ?
I would like to be sure before to start to compile it.

iericzhou:
As far as I know, this lib doesn't compile on Windows~
If any fellow succeeds in compiling in on Windows, please comment here.

xupanwilson:
I have build it on windows by visual studio 2013 successfully.
[...]

PR #106, opened 2016-11-10
CMake pure build on Windows

XeCycle:
Building on Windows has long been known to fail (#48), but if we skip the TCP part and do a pure build, it is easy to fix.  This adds a `-DPURE_BUILD=ON` option to CMake build, and fixes it on Windows, tested with VS 2015.  Personally I use the lib integrated into libuv, these should be enough for me to continue hacking.

Btw `tcpdefines.h` added by #80 was moved to `src/`; to me it looks like a simple internal header.  No idea whether you would like to keep it there (and have it installed).


src/includes.h
```diff
@@ -22,10 +22,14 @@
 #include <vector>
 #include <queue>
 #include <sys/types.h>
+
+#ifndef _WIN32
```
ricejasonf: Why not skip these for "BUILD_PURE" instead of just WIN32?

CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ if(BUILD_TUTORIALS)
 #    add_subdirectory(examples/rabbitmq_tutorials)
 endif()
 
+option(BUILD_PURE "build in pure mode" OFF)
```
ricejasonf: It would be a good idea to prefix `BUILD_PURE` with something unique. Like `AMQP_CPP_BUILD_PURE`.

XeCycle: I've seen projects following either way, so maybe it is a bit controversial here.  Some may feel it like `--with-tutorials` or `--without-tcp`, and others still are afraid of name conflicts.

Name conflicts come into play when this project is used by `add_subdirectory` from another project.  Looking at that set to parent scope, seems we are to support this use case?

Personally I prefer to prefix it, but here I'm just following existing code --- several lines above, `BUILD_TUTORIALS`.

src/includes.h
XeCycle: Good catch, will include that.

CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ if(BUILD_TUTORIALS)
 endif()
 
 option(BUILD_PURE "build in pure mode" OFF)
+if(BUILD_PURE)
+    add_definitions(AMQP_CPP_BUILD_PURE)
```
tretiy: Seems that AMQP_CPP_BUILD_PURE should have -D prefix

tommitytom:
What's the go with this?  Has it been merged in to the main repo?

LiamKarlMitchell:
So can we build on Windows VS2017?

surgura:
This has been fixed/sortof merged as of 91c4f891590c630df4f2a18e3f5925ddf41b3d3f, so we can close.

PR #115, opened 2017-02-16
build on windows with visual studio

fantastory:
Project for building with visual studio (https://github.com/fantastory/AMQP-CPP.nuget) has a submodule dependency.

- size_t is 32 bit on 32 bit windows builds
- windows functions here and there
- problems with operator[] while there exist implicit operator (char *)

### DIFF
--- a/amqpcpp.h
+++ b/amqpcpp.h
@@ -8,6 +8,10 @@
 
 #pragma once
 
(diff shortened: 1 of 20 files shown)


fantastory:
was done from master branch

PR #116, opened 2017-02-16
Vs2015 build

fantastory:
Project for building with visual studio (https://github.com/fantastory/AMQP-CPP.nuget) has a submodule dependency.

    size_t is 32 bit on 32 bit windows builds
    windows functions here and there
    problems with operator[] while there exist implicit operator (char *)

### DIFF
--- a/include/field.h
+++ b/include/field.h
@@ -87,7 +87,7 @@ public:
      *  @return mixed
      */
     virtual operator const std::string& () const;
-    //virtual operator const char * () const { return nullptr; } //can not use operaotr[] on subclasses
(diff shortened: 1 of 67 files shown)


include/field.h
```diff
@@ -87,7 +87,7 @@ class Field
      *  @return mixed
      */
     virtual operator const std::string& () const;
-    virtual operator const char * () const { return nullptr; }
+    //virtual operator const char * () const { return nullptr; } //can not use operaotr[] on subclasses
```
fantastory: i have no idea what to do with this :(

fantastory:
#48 boost example is working for in visual studio 2015

PR #133, opened 2017-06-07
table field value type extension

tom3135:
table's field value types extend 'x' for Qpid/Rabbit
the library rabbitmq-c has add "AMQP_FIELD_KIND_BYTES"('x')


EmielBruijntjes:
Is this type 'x' defined in the AMQP standard?

PR #139, opened 2017-07-02
fix linked list bug when you add more than one element to the list

rick-wildes:
I don't think more than one is ever added so that may be why this bug went unnoticed.  But if you ever added more than one element you would loose the old head.


EmielBruijntjes:
I don't see what is wrong with the old code. Did you run into a bug?

rick-wildes:
Never mind I got it backwards.

PR #145, opened 2017-09-19
Tweaked the directory structure to allow use in embedded builds

vs49688:
I needed to to have this as a subproject of another CMake project -- installation was not allowed.
This just tweaks the directory structure to allow this. Post-installation state is the same as before the change, so breakage should be minor.


EmielBruijntjes:
Im not going to merge this. I have never heard about the directory structure being a problem.

Issue #146, opened 2017-10-12
Implement boost::asio event handler


zerodefect:
At the moment, the AMQP-CPP library facilitates integration with various libraries that offer event loops. This not limited to the likes of libev and libuv. Nonetheless, I feel one of the more popular event handlers is missing - boost::asio.

On my short journey, I have stumbled upon various attempts at an event-driven boost::asio implementation that integrates with AMQP-CPP, but I've found they're not very clean and don't follow the general pattern of the existing event-driven implementations in this SDK. 

To try help myself (and others), I have spent some time writing a boost::asio implementation which I'm hoping that the maintainers will feel is fit for inclusion in this project.

Note: This implementation only works in POSIX-based systems. This is of no fault of my own (see [post here](https://github.com/chriskohlhoff/asio/issues/70)).  Let me explain. I try and wrap a file descriptor returned from AMQP-CPP in a TCP boost socket. This is all well while the socket is being watched; however, when it comes to a point where the boost socket needs to be disassociated with the file descriptor,  it is not possible on Windows based systems to do this without undesirable side effects. With this all being said, I don't feel a POSIX-only implementation should be overlooked.

zerodefect:
If this goes in, I'll look to add heartbeats (timer) and make some other tweaks. One thing I've realised is that I need to tackle the situation where the io_service is using multiple threads.

PR #147, opened 2017-10-12
Add boost asio handler

zerodefect:
As per issue #146 , this is an implementation of a boost::asio event handler that makes use of boost::asio::io_service.  On suitable for POSIX-based systems.


fantastory:
There was an AsioHandler in older copernica version have you seen it? It needed some fixes thought.
Why only POSIX? I can not find place where you disassociate sockets.

zerodefect:
@fantastory Thanks for the review and subsequent approval/merge. 

I had a look at the older AsioHandler, and I couldn't get it to work without hitting issues.  It's the reason that I started a new one from scratch.

The pertinent code which deals with disassociating from a socket:

https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/9559dfb880defdc39ffdc727b671c6835a6c1c58/include/libboostasio.h#L189-L197

The specific call is '_ _socket.release();_'

You can read more here - https://github.com/chriskohlhoff/asio/issues/70

PR #148, opened 2017-10-19
Add strand to boost handler

zerodefect:
Added a boost io::service_strand to LibBoostAsioHandler to make the AMQP-CPP handler thread safe when a boost::asio::io_service is used that is specifically utilising more than a single thread for its pool.

This code has been tested.


zerodefect:
Made one last commit since original creation of pull request. Added much needed example/test to 'test/libboostasio.cpp'.

Issue #162, opened 2018-01-17
Compile problem on test/libboostasio.cpp


Cereal84:
Hi,
I'm trying to compile 'test/libboostasio.cpp' using Makefile  but I've this problem:

/usr/include/amqpcpp/libboostasio.h: In member function 'void AMQP::LibBoostAsioHandler::Watcher::read_handler(const boost::system::error_code&, std::size_t, std::weak_ptr<AMQP::LibBoostAsioHandler::Watcher>, AMQP::TcpConnection*, int)':
/usr/include/amqpcpp/libboostasio.h:152:41: error: 'boost::placeholders' has not been declared
                                         STRAND_SOCKET_HANDLER(
......

I'm using libboost 1.58.0 .

billchenchina:
See #159. Same problem.

Cereal84:
My bad.

zerodefect:
I'll look to fix this in #159. Thanks.

zerodefect:
This is being resolved by another developer in #163.

PR #163, opened 2018-01-17
Fix for  boost example

Cereal84:
I've done some fixes about boost and its example.

### DIFF
--- a/include/libboostasio.h
+++ b/include/libboostasio.h
@@ -152,8 +152,8 @@ private:
                                         STRAND_SOCKET_HANDLER(
                                             boost::bind(&Watcher::read_handler,
                                             this,
-                                            _1,
-                                            _2,
+                                            boost::arg<1>(),
+                                            boost::arg<2>(),
 // C++17 has 'weak_from_this()' support.
 #if __cplusplus >= 201701L
                                             weak_from_this(),
@@ -197,8 +197,8 @@ private:
                                          STRAND_SOCKET_HANDLER(
                                              boost::bind(&Watcher::write_handler,
                                                          this,
-                                                         _1,
-                                                         _2,
+                                                         boost::arg<1>(),
+                                                         boost::arg<2>(),
 // C++17 has 'weak_from_this()' support.
 #if __cplusplus >= 201701L
                                                          weak_from_this(),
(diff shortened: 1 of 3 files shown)


zerodefect:
Thanks for making these changes. 👍  I had two (2) suggestions:

- We ought to make the necessary changes so that it builds with '-std=c++11' compilers. After all, this is a C++11 project.
- I would be hesitant to use _1 instead of the fully qualifying namespace. I would try the suggestion [here](https://svn.boost.org/trac10/ticket/11902).

On point #1, I can't  work out why it doesn't fail with my GCC compiler (v7.2). I can't reproduce any of these problems. Merely to increase my knowledge, could you tell me the build chain you're using?

zerodefect:
@EmielBruijntjes it would be extremely beneficial to have automated builds so that issues like this are caught earlier. Thoughts?

Cereal84:
Hi @zerodefect ,
about th 'C++11' it was my mistake, probably I confused some error of something else project which I'm working on. I've tried to use -std=c++11 and it works, so I will revert it to std=c++11.
About the second point I understand your thought but with boost-1.56 (on Gentoo) and boost-1.58 (on Ubuntu) I've the error described on #162 . 
I will try to investigate on the link you've suggested before.

Cereal84:
I've tried the suggestion described [https://svn.boost.org/trac10/ticket/11902](here) but it didn't work for me.

Cereal84:
At least on boost 1.56.0 (and I guess until 1.59.X) the in boost placeholders are in unnamed  namespaces, so it is not defined any boost::placeholders. 
@zerodefect,  Which version of Boost library do you have on your machine?

Cereal84:
Instead to use '_1/_2' or boost::placeholders (which in some version is not defined) I've used boost::argv<>(). 
This change resolve my problem and is explicited the namespace of boost instead on unnamed namespace like my old version.

zerodefect:
Nice work.

EmielBruijntjes:
@zerodefect At Copernica we have set up an automatic build framework for our own entire codebase, for which amqp-cpp is only a small part. But this does not run through github, but on our own development servers. So for me, there is no need for such a thing. (By the way: boost-asio examples are not included in this automatic build because we do not use that ourselves).

If you are however willing to set this up yourself and send in pull requests, that is fine by me.

Issue #164, opened 2018-01-23
clang/gcc compile error: embedding a directive within macro arguments...


lion10243:
Both clang/gcc throw errors compiling this library.

clang-3.9
error: embedding a directive within macro arguments has undefined behavior
 #if __cplusplus >= 201701L

gcc-5.4.0
error: embedding a directive within macro arguments is not portable
 #if __cplusplus >= 201701L

This problem occurrs because the `#if` directive is embedded within a preprocessor macro. The compiler error applies to all six lines including the __cplusplus directive check.

Also, the use of the macros (STRAND_XXX_HANDLER) should not be considered "good" modern cpp style (imho it's dirty ;-)

https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/9935f8414bdda7854bb43c18723fd56b1529a36e/include/libboostasio.h#L272

zerodefect:
Can we mark this as resolved and close please?

PR #165, opened 2018-01-23
Libboostasio fix

lion10243:
My suggested fix for issue #164 

I've also replaced the macros with short templated functions to avoid the preprocessor code gen.

### DIFF
--- a/include/libboostasio.h
+++ b/include/libboostasio.h
@@ -127,7 +127,13 @@ private:
          */
         handler_cb get_read_handler(TcpConnection *const connection, const int fd)
         {
-            auto fn = boost::bind(&Watcher::read_handler, this, _1, _2, PTR_FROM_THIS(), connection, fd);
+            auto fn = boost::bind(&Watcher::read_handler,
+                                  this,
+                                  _1,
+                                  _2,
+                                  PTR_FROM_THIS(),
+                                  connection,
+                                  fd);
             return get_dispatch_wrapper(fn);
         }
 
@@ -139,7 +145,13 @@ private:
          */
         handler_cb get_write_handler(TcpConnection *const connection, const int fd)
         {
-            auto fn = boost::bind(&Watcher::write_handler, this, _1, _2, PTR_FROM_THIS(), connection, fd);
+            auto fn = boost::bind(&Watcher::write_handler,
+                                  this,
+                                  _1,
+                                  _2,
+                                  PTR_FROM_THIS(),
+                                  connection,
+                                  fd);
             return get_dispatch_wrapper(fn);
         }
 
@@ -171,7 +183,9 @@ private:
 
                 _read_pending = true;
                 
-                _socket.async_read_some(boost::asio::null_buffers(), get_read_handler(connection, fd));
+                _socket.async_read_some(
+                    boost::asio::null_buffers(),
+                    get_read_handler(connection, fd));
             }
         }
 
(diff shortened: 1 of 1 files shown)


include/libboostasio.h
```diff
@@ -27,33 +26,35 @@
 #include <boost/bind.hpp>
 
 
-///////////////////////////////////////////////////////////////////
```
zerodefect: I'd wrap this in an anonymous namespace.

include/libboostasio.h
```diff
@@ -161,7 +160,13 @@ class LibBoostAsioHandler : public virtual TcpHandler
...
 #endif
                                             connection,
-                                            fd)));
+                                            fd);
+                _socket.async_read_some(
+                    boost::asio::null_buffers(),
+                    [f = func, strand = _strand](const boost::system::error_code &ec, const std::size_t bytes_transferred) {
+                        strand_sock_dispatch_func(f, strand, ec, bytes_transferred);
+                    }
+                );
```
zerodefect: We could minimise code duplication by adding a member function like the following:

```
std::function<void(boost::system::error_code,std::size_t)> get_io_handler(const bool bRead,
																		  TcpConnection *const connection,
																		  const int fd)
{
	auto func = boost::bind(bRead ? &Watcher::read_handler : &Watcher::write_handler,
							this,
							boost::arg<1>(),
							boost::arg<2>(),
// C++17 has 'weak_from_this()' support.
#if __cplusplus >= 201701L
							weak_from_this(),
#else
							shared_from_this(),
#endif
							connection,
							fd);
	return [f = func, strand = _strand](const boost::system::error_code &ec, const std::size_t bytes_transferred)
	{
		strand_sock_dispatch_func(f, strand, ec, bytes_transferred);
	};
}
```

...the you could just call:

`
_socket.async_write_some(boost::asio::null_buffers(), get_io_handler(true, connection, fd));
`

You should then be able to do something very similar for the timer

Note: Not sure if this compiles.

include/libboostasio.h
```diff
@@ -27,33 +26,35 @@
...
-    if (!apStrand)                                                                              \
-    {                                                                                           \
-        fn(boost::system::errc::make_error_code(boost::system::errc::operation_canceled),std::size_t{0}); \
-        return;                                                                                 \
-    }                                                                                           \
-                                                                                                \
-    apStrand->dispatch(boost::bind(fn,ec,bytes_transferred));                                   \
+template<class Func>
+void strand_sock_dispatch_func(
+    Func fn,
```
zerodefect: I'd make this `Func &&fn` and make use of move semantics.

zerodefect:
I'd suggest a few changes. I agree with the spirit of what you're trying to achieve though.



For the record, I'm not the gatekeeper around here, @EmielBruijntjes has the clout.

include/libboostasio.h
lion10243: True, that I will add.

include/libboostasio.h
lion10243: well spotted, thanks

include/libboostasio.h
lion10243: I like the suggestion - will see if it can be integrated.

lion10243:
@zerodefect  the updated version of the code is a bit cleaner and integrates the changes. The ternary bRead ? bind(read_handler,...) : bind(write_handler, ...)  wouldn't compile on either gcc or clang, so I wrapped the binding of read_handler and write_handler into separate, encasulating functions.

lion10243:
@EmielBruijntjes  Hello Emiel, my colleage and I, here in Berlin, have been users of the AMQP-CPP library for the past 1.5 years. With the recent 2.8.0 release we have unfortunately not been able to use the boost::asio handler in libboostasio.h out-of-the-box. Would you be willing to review the pull request, please, for inclusion in a future minor release of the library?

PR #167, opened 2018-01-23
Added continuous integration with travis.ci

zerodefect:
I've got the ball rolling here with enabling automated builds. There are a number of things missing which will need to be added in time:
 
- Building examples
- Creating unit tests and executing them (integrate something like GTest?)
- Updating CMakeLists.txt to a newer version

Hopefully, I've started something here which others can build upon. Note that the admin will need to enable automated builds using Github account through Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,17 +23,15 @@ matrix:
 
   - os: linux
     compiler: gcc
-    env:
-      - COMPILER=g++-5
+    env: COMPILER=g++-5
     addons:
       apt:
-        #sources: ['ubuntu-toolchain-r-test']
+        sources: ['ubuntu-toolchain-r-test']
         packages: ['g++-5', 'ninja-build']
 
   - os: linux
     compiler: gcc
-    env:
-      - COMPILER=g++-6
+    env: COMPILER=g++-6
     addons:
       apt:
         sources: ['ubuntu-toolchain-r-test']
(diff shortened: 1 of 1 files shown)


zerodefect:
Here is a summary of the build report(s) with my fork.

https://travis-ci.org/zerodefect/AMQP-CPP/builds

EmielBruijntjes:
Warning: AMQP-CPP works and we (Copernica) are happy with it. This library does not give us headaches so we can focus on other things. Our experience with external parties submitting test frameworks, example codes and other contributions to our open source projects _"to make things better"_, but that do not add new features or fix bugs, have not always been good: these "improvements" often only cause a lot of noise and take time. It this travis-thing is going to do the same, I will not hesitate to deactivate it again.

zerodefect:
Thanks for your feedback. It's disappointing to hear of your bad experiences (note: I'm not blaming you for that).

Are you able to elaborate further on some of your past experiences with these sorts of tools. Are you talking about false positives or system failures (ie. Problems independent of the code itself)?

From my point of view, it gives me reassurance knowing that the code builds and has had some preliminary testing. Message buses are often fundamental to 
software. I appreciate though that I'm not the only one that counts.

Feel free to send me an email if you prefer... gavin(dot)smith(at)coralbay(dot)tv

PR #170, opened 2018-01-29
Add Examples (2) and package file info

Cereal84:
Hi,
I've add two examples of library use (for libboost only) like the first 2 in rabbitmq tutorial.
Also for UNIX system add package info file, really useful for autotools and to retrieve informations usign pkg-config --libs amqpcpp.

I hope you appreciate it.


EmielBruijntjes:
Do we still need this after all changes to the makefiles?

Cereal84:
Makefiles do not generate pgk-config file (useful on Linux like system) and examples can be useful for who use the library.

surgura:
What is the purpose of the changes to `include/libboostasio.h`?

Cereal84:
@surgura  probably is a conflict due the fact that I have not merged the current master.

Cereal84:
Anyway I messed up remerging another branch of mine in my master, there are other changes about version number retrived directly from tag version. In this way there is no problem to change version number.

wosrediinanatour:
This would be cool to be in the next release...

PR #172, opened 2018-01-30
Added missing include for "boost/function.hpp" in libboostasio.h.

svenslaggare:
The 'libboostasio.h' file requires boost::function but is not imported directly in the file.


EmielBruijntjes:
Thanks. I've "merged" this by hand.

PR #173, opened 2018-01-30
Build examples now with travis-ci to ensure new changes 

zerodefect:
…do not break existing code.

Few changes here but nothing too significant:

- Use Docker containers to build code because travic-ci uses OLD ubuntu flavour (Trusty). With Docker now able to use Artful.
- Build code examples. Moved applicable code examples into examples directory. Admittedly, not updated Makefile. Need some help here 👍 
- Fixed warnings in libboostasio.h because of use of c++14 lambda capture initializers.


surgura:
``` C++
if(AMQP-CPP_BUILD_EXAMPLES)
    add_subdirectory(examples)
endif()
```
Could you move this up to just above
``` CMake 
# settings for specific compilers
# -------------------------------------
```
That way we have it in on place with the our sources. Right now you have it in the part about installing.

Is there a reason you include all .h files in the CMakelists? Earlier I removed them, because you only compile .cpp anyway.

zerodefect:
> Is there a reason you include all .h files in the CMakelists? Earlier I removed them, because you only compile .cpp anyway.

Yes. Header files are still part of the project (this is regardless of how the compiler handles them).

zerodefect:
> Could you move this up to just above

Done.

Issue #179, opened 2018-01-31
travis gcc-5 support missing/removed


lion10243:
Very recently GCC-5 support was removed from the travis configuration, which I believe was related to compile errors (e.g. "to_string not being a member of std::")

As a library user, we are still using gcc-5 (we cannot upgrade to gcc 6/7 just yet) and it would nice if travis could regularly test a gcc-5 build.

lion10243:
PR 180 added

PR #180, opened 2018-01-31
support travis gcc 5

lion10243:
PR for issue #179 

Hello @zerodefect  @surgura ,  I would like to re-add travis support for gcc5, which was removed recently, I believe due to compile errors.  Would that be a problem?

I have tested the build in travis using a private branch, and the build errors that occurred can be fixed by passing the compile flag ''-std=c++11".

I have also corrected the cmake syntax in travis.yaml inline with cmake --help (correct syntax is cmake [options] <path>)

Feedback welcome.


zerodefect:
Looks good. Thanks for fixing!!  

It was on my radar, but I've not had time to get onto it.  It seems that when I moved from Trusty to Artful in Docker, it wouldn't work out the box.  I was desperate for CI for further changes before people broke libboostasio.h again, so I was going to come back to it :)

Issue #181, opened 2018-01-31
Customise the rabbitmq heartbeat interval


lion10243:
We have a use-case with the library where we want to explicitly negotiate the heartbeat interval with rabbitmq to, e.g., 60 seconds.  Currently, the library (in my case, libboostasio handler) automatically accepts whatever time interval rabbitmq sends.

I would like to provide a custom "uint16_t timeout" value for the heartbeat interval.

Using the libboostasio handler, I cannot find a way to specify the heartbeat timeout, without providing a custom constructor.

If there is already a way to do this in the library, please let me know.

PR #182, opened 2018-02-01
add second constructor with heartbeat interval param

lion10243:
A PR for issue #181 

A second ctor in libboostasio was added to control the heartbeat interval.
Applied modern-style cpp typedefs (using).

New ctor signature:
explicit LibBoostAsioHandler(boost::asio::io_service &io_service, uint16_t interval)

@EmielBruijntjes  @zerodefect  do you both know if there is an alternative method for configuring the heartbeat in the library, without a requiring a new constructor?

Thanks in advance.


EmielBruijntjes:
Extend from the ConnectionHandler class and implement your own onNegotiate() method.

lion10243:
@EmielBruijntjes  Thanks for the suggestion, that worked.
I've updated the PR to change the class access modifier in LibBoostAsioHandler from 'private' to 'protected' to allow the class to be extended.

PR #189, opened 2018-02-22
Add returned message support.

RZRCDR:
Added support to handle returned messages.
These are messages that are published with the mandatory or immediate flag set.
If any of these message are unroutable they will be returned to the client and processed.

A new callback type called ReturnCallback has been added for these messages.
This callback gets registered on the AMQP::Channel via onReturn()

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,15 +803,13 @@ in almost any form:
  *  The following flags can be used
  *
  *      -   mandatory   if set, an unroutable message will be sent back to
- *                      the client (currently not supported)
+ *                      the client
  *
  *      -   immediate   if set, a message that could not immediately be consumed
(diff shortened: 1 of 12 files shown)


EmielBruijntjes:
Thanks for your contribution. I've used it as inspiration it to support the following way of handling returned messages:

```c++
// callback that handles messages that could not be sent to a queue
auto callback = [](const AMQP::Message &message, uint16_t code, const std::string &descripton) {
    std::cout << "message could not be published: it was returned" << std::endl;
};

// publish a message, and redirect them to a callback if they cannot be sent to a queue
channel.publish("exchange", "routingkey", "message", AMQP::mandatory).onReturned(callback);
````

PR #192, opened 2018-03-02
Fix build_shared - EXPORT keyword in wrong position

lion10243:


lion10243:
@EmielBruijntjes this PR fixes a bug in CMake when BUILD_SHARED=ON

PR #195, opened 2018-03-06
Fixed MinGW compilation on Windows

Aleksandrius:
`ntohll` function is missing in `winsock2.h` on MinGW installation. Although the official Microsoft SDK has an implementation of `ntohll` function, I replaced it with the same implementation based on `ntohl`. Previosly, someone did the same job for `htonll` function.


surgura:
I did the one for htonll. Did not have compilations issues, but definitely have reason to believe some systems have. I pulled mine from here https://stackoverflow.com/a/28592202/2476219. ntohll is in the same answer as well, so I'm fairly sure this is correct.

PR #201, opened 2018-03-16
Simplify inclusion as a subproject

i-rinat:
The patch changes build scripts to use not root, but current directories, which simplifies using this project as a CMake subproject. When included somewhere, `${CMAKE_SOURCE_DIR}` variables points to the source directory of a parent project, while `${CMAKE_CURRENT_SOURCE_DIR}` still points to the right place.

Another thing is installing generated library to the source directory. Having a separate directory makes it easier to spot the generated output, since it's not polluted by CMake-generated files. But it's still appears in a sources directory, which may surprise. The patch proposes to move it to the `build/bin/`. It's still a clean subdirectory, but at the same moment is located in build subtree, where it's expected the most.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ set (CMAKE_CXX_STANDARD 11)
 # ------------------------------------------------------------------------------------------------------
 
 # set include/ as include directory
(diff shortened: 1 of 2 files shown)


surgura:
This is fine by me.

PR #206, opened 2018-04-01
Add method to return the amount of queued outgoing data

MikePlayle:
I have a program which needs to publish a large number of messages and then exit.

Exiting after calling AMQP::Channel::publish() for the last time doesn't work, because most of the messages are still queued within libamqpcpp. I need to wait for them all to go out on the wire before shutting down.

If I patch my program so that it never exits, all the messages eventually reach RabbitMQ and the rest of the system works perfectly. However my program's peak memory consumption is enormous.

This change adds the TcpConnection::bytesQueued() method which returns the number of bytes currently waiting to be sent across the TCP connection. Monitoring its return value solves both of my problems, and might be useful for others too.

### DIFF
--- a/include/amqpcpp/linux_tcp/tcpconnection.h
+++ b/include/amqpcpp/linux_tcp/tcpconnection.h
@@ -123,6 +123,12 @@ public:
     int fileno() const;
 
     /**
+     *  The number of outgoing bytes queued on this connection.
+     *  @return size_t
+     */
+    size_t bytesQueued() const;
+
+    /**
      *  Process the TCP connection
      * 
      *  This method should be called when the filedescriptor that is registered
--- a/src/linux_tcp/sslconnected.h
+++ b/src/linux_tcp/sslconnected.h
@@ -335,6 +335,12 @@ public:
      *  @return int
      */
     virtual int fileno() const override { return _socket; }
+
+    /**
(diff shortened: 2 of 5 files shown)


EmielBruijntjes:
I renamed the method from `bytesQueued()` into `queued()`, because other methods that return a size also do not have the "bytes" word into their names (like TcpConnection::expected()". I also added a couple of "override" keywords here and there.

Wit your use case I can foresee that you also want to be notified when the queue is gone, and the connection is available again for direct writing. If you need this too, you will have to send in another pull request with a couple of extra methods in the tcp-handler that get called when the connections starts buffering / unbuffering / becomes unbuffered.

Keep in mind that AMQP-CPP can also maintain queues on a per-channel basis. You may have to check that as well.

Instead of running your program foreverer to be sure that everything is published, you can also close the connection after sending the last message, and wait for your custom connection-handler "onClosed()" handler to be called.

MikePlayle:
Thanks for the pointer to `onClosed()`. Somehow I'd missed this, and it does indeed improve my end-of-connection handling enormously compared to just destroying the object - though the `queued()` method is still needed to avoid filling enormous amounts of RAM with buffered messages.
Right now I don't foresee a need for any additional notifications; if they become important in the future, I'll add them them.

PR #216, opened 2018-04-22
Fix MSVC compile warnings

RZRCDR:
Fix MSVC compile warning **C4800** for bitwise operation resulting in int to boolean assignment.
Fix MSVC compile warning **C4244** possible loss of data. Possible data loss from uint64_t to uint32_t.

### DIFF
--- a/include/amqpcpp/callbacks.h
+++ b/include/amqpcpp/callbacks.h
@@ -45,7 +45,7 @@ using DeleteCallback        =   std::function<void(uint32_t deletedmessages)>;
  *  When retrieving the size of a queue in some way
  */
 using EmptyCallback         =   std::function<void()>;
-using SizeCallback          =   std::function<void(uint32_t messagecount)>;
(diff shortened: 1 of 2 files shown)


include/amqpcpp/callbacks.h
```diff
@@ -45,7 +45,7 @@ using DeleteCallback        =   std::function<void(uint32_t deletedmessages)>;
  *  When retrieving the size of a queue in some way
  */
 using EmptyCallback         =   std::function<void()>;
-using SizeCallback          =   std::function<void(uint32_t messagecount)>;
+using SizeCallback          =   std::function<void(uint64_t messagecount)>;
```
pabigot: I'm wondering whether it would be worth it to add a CountCallback that retains the current signature and use that as the type of `_countCallback` in DeferredGet which is returning a value that is uint32_t by protocol.  Most other uses of `SizeCallback` are providing message size which does need to be uint64_t.

PR #221, opened 2018-04-26
Support client certificate authentication

trevorperrin:
Modified the openssl wrapper to add functionality for authenticating
client certificates.  Added to the example set for how to use certain
functions.


EmielBruijntjes:
I'm not so sure about this. OpenSSL has a very long list of functions, and I do not want to add every single one of them to the AMQP-CPP library. 

The TcpHandler::onSecured() method is called when the connection has been setup and can be overridden in user space to check if the connection is secure enough for your liking. Is this interface currently not powerful enough for you? Maybe we need some extra virtual methods in the TcpHandler class _before_ the connection is set up to allow applications to further initialize the SSL structure?

trevorperrin:
I need to do both server and client authentication, and the call to onSecured() in the handler occurs too late in the process (after the TLS handshake).  This means the client will need a key/cert pair during the handshake, and be able to load a CA certificate for verification purposes.  It also needs access to set the verification mode, supply the callback, load the peer cert, and get the verification results.

Currently, I have supplied these functions via the OpenSSL wrapper (which seemed like the logical place), and am using this functionality at the first point in which it would be available for use - right after the SSL object is created in the SslWrapper.  This follows the example left in the comment in the SslWrapper constructor.  The LibEvHandler onSecured() method has been overridden to get the peer certificate and then check the verify result using the SSL object.

Somehow the application will need access to these lower level OpenSSL functions.  Whether that functionality is supplied via the OpenSSL wrapper itself, or via a modified handler, is a good question.  Having the ability to initialize the SSL structure from outside the library, before the TLS handshake occurs, would be of value.

EmielBruijntjes:
I am not going to merge this one. But what could be interesting, is to add an additional method to the TcpHandler class that allow user space code to create and initialize their own SSL_CTX and SSL structures. This would give maximum flexibility to setup secure connection to your own liking:

````c++
/**
 *  Method that is called by AMQP-CPP that allows user space applications to create
 *  and initialize their own openssl SSL* structure that is going to be used to secure
 *  the connection. The default implementation simply calls the SSL_CTX_new() and 
 *  SSL_new() functions, but applications that want to (for example) share contexts 
 *  between connections, or add their own initialization routinges, can override this
 *  method and do their own initialization.
 *  @param  connection   The connection to be initialized
 *  @return SSL*         The SSL* structure to be used
 */
SSL *TcpHandler::onSecure(TcpConnection *connection)
````

Do you think this would be useful, and is it possible to send in a pull request for this?

EmielBruijntjes:
Not merging this one, waiting for a better solutions.

PR #222, opened 2018-04-26
Clear SSL Error Queue

trevorperrin:
The OpenSSL documentation states that the error queue for the current
thread needs to be emptied prior to the next TLS/SSL I/O operation,
or the call to SSL_get_error() will not be reliable.


EmielBruijntjes:
Besides your changes, I added a couple of extra calls to other operations as well, and I relocated some of our calls. Could you please have a look to see if it satisfies openssl's requirements?

PR #224, opened 2018-05-02
Add OnCreateSSL to TcpHandler

trevorperrin:
This change adds a function to the TcpHandler which allows the user to initialize the SSL structure prior to the start of the handshake.


EmielBruijntjes:
Why exactly do you need this feature, what sort of extra initialization do you need that is not yet offered by AMQP-CPP? I am a little worried that there might be a conflict in how the user initializes the SSL structure, and how AMQP-CPP does this. 

And do you think that is might be necessary to allow the user to pass in a SSL_CTX and/or SSL structure instead of allocating and initializing this from inside AMQP-CPP?

troian:
Looks like we trying to do same
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/285
@EmielBruijntjes allowing to initialize SSL object before handshake is quite important is for example it allows to set SSL certificates and key for encryption and authentication. Current API simply does not allow to do that.

henris42:
Is there any news on implementing the client certificate support feature? For my project that is a MUST requirement, and I'd hate to maintain a fork.

EmielBruijntjes:
This has been implemented in 3dbd045d0601e6ce9d8b847223a22e08ac422575

However, would it not be more helpful to add support for client side certificates via parameters to the address. For example to support addresses like "amqps://hostname/vhost?certificate=/path/to/file"?

henris42:
Great! The certificate input side needs to be very flexible, for example we have our own PKI systems and we do not store certs or keys in plain files at all. We decrypt keys from keystore only when needed. 
It would be useful to have a nicer API to feed the keys and certs at login phase though. OpenSSL however has some ugly limitations there, and I cannot see we can get rid of custom code for a while.

PR #228, opened 2018-05-29
Restructuring the SSL event handling

### LINKED ISSUE #207: Failed AMQP handshake with TLS
I'm implementing a client which connects via TLS (using the software versions listed below).  During testing, the TLS handshake succeeds, but the AMQP portion of the handshake fails.  While attempting to establish the connection, the AMQP header ("AMQP0091") is successfully sent by the client, and the server replies with a Connection.Start message.  At this point, the client is supposed to reply with a Connection.Start-Ok message, but it does not do so.

The data is being generated (connectionstartframe.h), and the connection is set to send the ConnectionStartOKFrame.  I followed the path through the following:

ConnectionStartFrame::process
ConnectionImpl::send
PassthroughBuffer::flush
TcpConnection::onData
SslConnected::send

In SslConnected::send, _handler->monitor is called with (readable | writable), but the write never occurs and the connection times out.  

The workaround I've used is to call write in the SslConnected::send function, right before the call to _handler->monitor:

`virtual void send(const char *buffer, size_t size) override`
`{`
`    ...`
`    Monitor monitor(this);`
`    write(monitor); // flush(monitor) works as well`
`    // comment...`
`    _handler->monitor(_connection, _socket, readable | writable);`
`}`

This "works"; the AMQP handshake succeeds, and data is sent and received in the client via TLS.  I don't like this solution, as I assume the Monitor should be tracking the Connection object, not SslConnected.  I'm also concerned about edge cases around the handling of the output buffer, and SSL error messages like SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE.  I have a solution that adds a monitor object to the send call of every TcpState object, in the event that any of those state objects need to monitor the connection.  If this sounds reasonable, I can issue a pull request.

What I'm using:
OS Version: CentOS 7.4
AMQP-CPP Version: AMQP-CPP Master (as of April 2, 2018)
openssl Version: 1.0.2k
RabbitMQ Version: 3.7.2
Event Handler: libboostasio.h

tilsche:
As per the OpenSSL documentation:

> Any TLS/SSL I/O function can lead to either of SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE.  In particular, SSL_read() or SSL_peek() may want to write data and
           SSL_write() may want to read data.

The previous implementation considered this only partially and inconsistently. Particularly a `write` could call a `read` when it was caused by a `readable` event, but then the `read` failed and changed the state to `reading` and no more writing was done.

This PR simplifies the whole read/write handling. Basically the first read/write is always attempted¹ to find out what SSL wants (`SSL_ERROR_WANT_READ/WRITE`). This is saved in the state and used for fd monitoring. For the next `process()`, these flags are used as filters. `read`, `write`, and `proceed` is always invoked directly from `process`.

¹ The constructor does not attempt read/write, it sets the `want_flags` to `readable | writable` instead to watch and react to any kind of fd events.

Fixes #207 . This is only weakly tested with ASIO and libev. It needs more test & review.

Thanks to @bmario @kinnarr

### DIFF
--- a/examples/libboostasio.cpp
+++ b/examples/libboostasio.cpp
@@ -19,6 +19,9 @@
 #include <amqpcpp.h>
 #include <amqpcpp/libboostasio.h>
 
+#include <openssl/ssl.h>
+#include <openssl/opensslv.h>
+
 /**
  *  Main program
  *  @return int
@@ -32,9 +35,16 @@ int main()
 
     // handler for libev
     AMQP::LibBoostAsioHandler handler(service);
-    
+
+    // init the SSL library
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+    SSL_library_init();
+#else
+    OPENSSL_init_ssl(0, NULL);
+#endif
+
     // make a connection
-    AMQP::TcpConnection connection(&handler, AMQP::Address("amqp://guest:guest@localhost/"));
+    AMQP::TcpConnection connection(&handler, AMQP::Address("amqps://guest:guest@localhost/"));
     
     // we need a channel too
     AMQP::TcpChannel channel(&connection);
@@ -44,7 +54,6 @@ int main()
         
         // report the name of the temporary queue
         std::cout << "declared queue " << name << std::endl;
-        
         // now we can close the connection
         connection.close();
     });
--- a/examples/libev.cpp
+++ b/examples/libev.cpp
@@ -144,10 +144,11 @@ int main()
 #endif
 
     // make a connection
-    AMQP::Address address("amqp://guest:guest@localhost/");
-//    AMQP::Address address("amqps://guest:guest@localhost/");
+//    AMQP::Address address("amqp://guest:guest@localhost/");
+    AMQP::Address address("amqps://guest:guest@localhost/");
     AMQP::TcpConnection connection(&handler, address);
     
+
     // we need a channel too
     AMQP::TcpChannel channel(&connection);
 
--- a/src/linux_tcp/sslconnected.h
+++ b/src/linux_tcp/sslconnected.h
@@ -5,8 +5,8 @@
  *
  * @copyright 2018 copernica BV
  */
- 
-/** 
+
+/**
   * Include guard
   */
 #pragma once
@@ -16,9 +16,23 @@
  */
 #include "tcpoutbuffer.h"
 #include "tcpinbuffer.h"
+#include "tcpstate.h"
 #include "wait.h"
 #include "sslwrapper.h"
 #include "sslshutdown.h"
+
+#include <amqpcpp/monitor.h>
+#include <amqpcpp/flags.h>
+
+#include <amqpcpp/linux_tcp/tcpconnection.h>
+#include <amqpcpp/linux_tcp/tcphandler.h>
+
+#include <cassert>
+
+extern "C"
+{
+#include <unistd.h>
+}
 
 /**
  * Set up namespace
--- a/src/linux_tcp/sslconnected.h
+++ b/src/linux_tcp/sslconnected.h
@@ -39,10 +39,10 @@ extern "C"
  */
 namespace AMQP {
 
-/** 
+/**
  * Class definition
- */ 
-class SslConnected : public TcpState, private Watchable 
+ */
+class SslConnected : public TcpState, private Watchable
 {
 private:
     /**
@@ -51,12 +51,12 @@ private:
      */
     SslWrapper _ssl;
 
-    /** 
+    /**
      *  Socket file descriptor
      *  @var int
      */
     int _socket;
-     
+
     /**
      *  The outgoing buffer
      *  @var TcpBuffer
@@ -68,23 +68,13 @@ private:
      *  @var TcpInBuffer
      */
     TcpInBuffer _in;
-    
-    /**
-     *  Are we now busy with sending or receiving?
-     *  @var enum
-     */
-    enum State {
-        state_idle,
-        state_sending,
-        state_receiving
-    } _state;
-    
+
     /**
      *  Should we close the connection after we've finished all operations?
      *  @var bool
      */
     bool _closed = false;
-    
+
     /**
      *  Have we reported the final instruction to the user?
      *  @var bool
@@ -96,7 +86,20 @@ private:
      *  @var size_t
      */
     size_t _reallocate = 0;
-    
+
+    /**
+     *  Remember what the last logical SSL write operation that failed wanted from the socket
+     *  @var int
+     */
+    int _write_want_flags;
+
+    /**
+     *  Remember what the last logical SSL read operation that failed wanted from the socket
+     *  This is initialized to both readable and writable to ensure the first process() will
+     *  try to read.
+     *  @var int
+     */
+    int _receive_want_flags = readable | writable;
 
     /**
      *  Close the connection
@@ -106,16 +109,16 @@ private:
     {
         // do nothing if already closed
         if (_socket < 0) return false;
-        
+
         // and stop monitoring it
         _handler->monitor(_connection, _socket, 0);
 
         // close the socket
         ::close(_socket);
-        
+
         // forget filedescriptor
         _socket = -1;
-        
+
         // done
         return true;
     }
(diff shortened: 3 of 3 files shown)


tilsche:
Could we maybe get some feedback to this PR? I'm also wondering how this relates to 80ce6327bb1e16472bf461f1561342f5cb454b2a

tilsche:
We have not tested 80ce632 because we have been running this RP's code-base for the last 5 months.

I don't think 80ce632 covers all problematic cases. The state handling is still messy. Consider the example that we observed failing:

> Particularly a write could call a read when it was caused by a readable event, but then the read failed and changed the state to reading and no more writing was done.

I can see this still happening as there is no guarantee that the read will fully succeed just because the socket was actually readable (as opposed to wrongly cached readable).

Overall I think running your own checks on the socket is the wrong way to go if you are in an async loop. This introduces possibilities for and redundancy, as well as many redundant syscalls and thus performance issues.

That said the blocking `select` call in `flush` should probably seem highly problematic to me as well, but that may be another topic...

tilsche:
@EmielBruijntjes We took a look at the recent b81bc340b541e43cb2f98aa28705bd0abb815b1e. It seems that it still uses the limited states which can end up inconsistent.

While reproducing issues with this is extremely difficult, I believe there is a clear argument for the code. We are happy and would appreciate to discuss this further.

EmielBruijntjes:
I do agree that the earlier implementation of the SslConnected class was not correct. But since the latest changes I have the impression that it does work. Can you explain which (edge) cases we are missing with the current implementation?

tilsche:
The case I am concerned about is the following:

1. SSLConnected::write
2. _out.sendto() returns > 0
3. still data in _out
4. socket becomes readable because of handshake data coming in
5. falling through the loop because of isReadable()
6. SSLConnected::receive
7. receive will handle the handshake but fails with SSL_ERROR_WANT_READ because there is no actual payload data to receive
8. _state = state_receiving
9. onIdle(readable)
10. we could write now, but process is never called
11. even if process would be called we would be stuck in _state == state_receiving which fails with WANT_READ as long as the server doesn't send anything, maybe never

EmielBruijntjes:
About point 7: does this indeed happen? I though that SSL_ERROR_WANT_READ is only returned when the read-operation should be repeated. If the read-operation fails because no data is available, it returns 0 and goes back to the idle state.

EmielBruijntjes:
About step 10: is that permitted? If SSL_ERROR_WANT_READ was returned by an earlier call, is it then allowed to call ssl_write() instead of repeating the ssl_read() operation?

EmielBruijntjes:
I don't think it is permitted to mix ssl_read() and ssl_write() calls if one of them returned WANT_READ or WANT_WRITE. As far as I see, the only option then is to repeat the call. The scenario that you are painting is, simplified, as follows:

1. socket becomes readable because of handshake data coming in.
2. we call ssl_read() because of this.
3. this returns SSL_WANT_READ because the data was only handshake data and there is no actual data.
4. from now on the only legal method to call is ssl_read() because openssl does not allow calls to ssl_write() or ssl_shutdown() until the previous ssl_read() is finished.
5. the program will never write again.

This is either an (unlikely) bug in openssl, or openssl already deals with this and this scenario does not occur.

tilsche:
I think the StackOverflow answer is wrong, but there is also a link to the [mailing list](https://marc.info/?l=openssl-dev&m=124312461512355&w=2
) which discusses a slight variation scenario I am concerned about:

> 1) You call SSL_write. It returns -1, WANT_READ, because a renegotiation is
> in progress.
> 
> 1A) The renegotiation data arrives.
> 
> 2) Independently, you SSL_read. It reads the renegotiation data but no
> application data, so it returns WANT_READ.
> 
> 3) You select for read, since OpenSSL told you to do so. The select will
> take forever, because the renegotiation data has already arrived and been
> read. Your SSL_write would succeed immediately now.
> 
> When SSL_read returns *anything* it invalidates any prior WANT_* return from
> SSL_write because the SSL_read may have made forward progress. Any call to
> an SSL_* function may consume data. You cannot assume that because the data
> hadn't been read when you called SSL_read it still hasn't been read after
> you called SSL_write.
> 
> This will screw you on renegotiations.

In AMQP-CPP, step 1) would cause a repeated `SSL_write` instead of `SSL_read`. But I don't see any reason why this 1A) could not happen without 1). The mailing list post strongly indicates that mixing `SSL_read`/`SSL_write` is ok.

One point regarding this PR is that the want-flags should be reset to their default whenever any of the functions has received new want flags. I somehow think the default want-flags for `SSL_write` should be only writable.

Unfortunately the official documentation is indeed not very helpful. There is one paragraph in `man SSL_get_error`

> It is safe to call SSL_read() or SSL_read_ex() when more data is available even when the call that set this error was an SSL_write() or SSL_write_ex().  However if the call
> was an SSL_write() or SSL_write_ex(), it should be called again to continue sending the application data.

The second sentence is rather confusing - I suspect it means "it should *eventually* be called again". It doesn't say whether it is safe to call `SSL_write` after a `SSL_read` failed.

tilsche:
There's also this paragraph in `man SSL_get_error` (emphasis mine):

> Caveat: Any TLS/SSL I/O function can lead to either of SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE.  In particular, SSL_read_ex(), SSL_read(), SSL_peek_ex(), or SSL_peek()
> may want to write data and SSL_write() or SSL_write_ex() may want to read data.  This is **mainly** because TLS/SSL handshakes may occur at any time during the protocol
> (**initiated by either the client or the server**); SSL_read_ex(), SSL_read(), SSL_peek_ex(), SSL_peek(), SSL_write_ex(), and SSL_write() **will handle any pending handshakes**.

One might read into this

- Server starting a handshake can very well happen, and thus my "failed read - never write" scenario
- The listed functions all handle pending handshakes, thus you may mix them - and should reset the want-flags from previous calls to their presumed defaults.

EmielBruijntjes:
Thanks for all your input. I find it very interesting because it would mean that I made some mistakes in other software too. But there are still a lot of unanswered questions here:

1. Is it ok to mix ssl_read and ssl_write? I always read the documentation so that the _same call_ had to be repeated. The documentation that you quote does indeed say that it is ok to start writing in the middle of a read-operation, but not the other way around (at least that is how I read it). And you are quoting openssl 1.1, while we want to support openssl 1.0 too, which does not contain this text. Maybe openssl 1.1 is more tolerant than 1.0?

2. But even if it is not forbidden to mix them, is it then also _wrong_ to not do that, and repeat the same operation until it is finished, before switching to the other operation? My opinion is that this is not wrong.

3. Does it ever happen that a peer starts a renegotiation handshake without it being part of a normal data transfer or shutdown? In other words: there is no intention to send data, but still the peer sends a SSL handshake? This is the 1A scenario. This would indeed trigger a ssl_read, that could indeed lead to an infinite want-read situation. But I doubt this happens. I _think_ that the handshakes only take place when some meaningful data is sent over the line too (afaik openssl does not have an internal timer that spontaneously start sending data over the line, the other party must thus have called ssl_write() or ssl_shutdown() to trigger the handshake). But it should not be so difficult to write a small server application that sends no data, but that does periodically send renegotiating handshakes? That would be the perfect demonstration that our SslConnected class is conceptually wrong.

tilsche:
I don't have authoritative answers, just my interpretation of the documentation / mailing list.

1. I did not find anything that says that you can't. Nothing says that you must repeat an operation before calling the other. For what it's worth`SSL_read` says (emphasis mine) "The calling process then **must** repeat the call after taking appropriate action to satisfy the needs of the read function", but `SSL_get_error` says  "**can** be retried later", "... **can** be called again". I don't know about 1.1, to some extent I am assuming that it is just refinement in the documentation.

2. Well - if you need to send data you must call `SSL_write` - in that sense it's wrong to not call it. I think the problem with AMQP is that it's asynchronous whereas with most other protocols its always clear who needs to send and who needs to receive next.

3. I read `SSL_get_error` that way: "TLS/SSL handshakes may occur at any time during the protocol          (initiated by either the client or the server)".

EmielBruijntjes:
My colleague has been further looking into this. We now think that it could indeed be possible to mix calls to ssl_read() and ssl_write(), but we do have to check it more closely to be sure. We also discovered that although the openssl 1.0 docs says that a repeated ssl_write() must be called with _exactly the same parameters_ as the previous call (our interpretation: with the same memory addresses), that the call can in reality also be repeated with the same data, but stored on a different location. This would allow us to copy buffers only when a repeat is necessary, instead of prematurely before each call just in case the call fails.

Anyway, there is some room for improvement in the SSL handling, because we no longer have to delay writing if a read operation is also in progress, and the other way around. And we do not have to buffer all outgoing data, but only the data that could not be sent at the first attempt. Whether the current implementation is now also _broken_, is not yet sure.

Issue #208, opened 2018-04-12
build requires openssl to be installed on build machine in /usr/include


yuyongzhang1:
Hi,

We use meson/ninja as our build systems, where openssl is built as 3rdParty package, which is not installed on build-machine.

We tried to build AMQP-CPP as another 3rdParty Package with conanfile as the following:
(We want to use TcpHandler, so we enabled AMQP-CPP_LINUX_TCP flag.)
===============================
from conans import ConanFile, CMake

class amqpcppConan(ConanFile):
    name = "amqpcpp"
    version = "3.0.0"
    settings = "os", "compiler", "build_type", "arch"
    options = {"shared": [True, False]}
    default_options = "shared=False"
    requires = "OpenSSL/1.0.2o@casa/stable"
    url = "https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git"
    license = "Apache License 2.0"
    description = "amqpcpp"

    def source(self):
        self.run("git clone https://github.com/CopernicaMarketingSoftware/AMQP-CPP.git")
        self.run("cd AMQP-CPP && git checkout v%s" % self.version)

    def build(self):
        cmake = CMake(self, generator='Ninja')
        #cmake.configure(source_folder="AMQP-CPP")
        cmake.configure(source_folder="AMQP-CPP",
             defs={"AMQP-CPP_LINUX_TCP" : "ON"})
        cmake.build()

    def package(self):
        self.copy("*.h", dst="include", src="AMQP-CPP/include")
        self.copy("*.a", dst="lib", keep_path=False)
        self.copy("*.so*", dst="lib", keep_path=False, symlinks=True)

    def package_info(self):
        self.cpp_info.libs = ["amqpcpp"]
=====================================

However, we are getting errors:
================================================
[2/17] Building CXX object CMakeFiles/amqpcpp.dir/src/linux_tcp/tcpconnection.cpp.o
FAILED: CMakeFiles/amqpcpp.dir/src/linux_tcp/tcpconnection.cpp.o 
[...]

PR #234, opened 2018-06-24
RFC: meson: add mesonbuild support

pabigot:
I recognize there's enough complexity trying to maintain both CMake and Makefile, but [meson]( https://mesonbuild.com/) has a few benefits including trivial support for Google Test and doxygen, and better provision of dependencies in pkgconfig than CMake.

Using AMQP-CPP as a subproject in a meson-based application becomes a simple matter of dropping something like the following as `subprojects/amqpcpp.wrap`:
```
[wrap-git]
directory = amqpcpp
url = /mnt/devel/external/RabbitMQ/AMQP-CPP
revision = dd00810f7fa3cc88470c52b274ea65c333cb5d66
```
and adding this to the project `meson.build`:
```
amqp_dep = dependency('amqpcpp',
                      default_options: [
                        'linux-tcp=true',
                      ],
                      fallback: ['amqpcpp', 'amqpcpp_dep'])
```

I've only tested this on Linux and with linux-tcp set.

If this is something that could be merged, I'd be happy to add some documentation.  The primary value to this project of merging it is Google Test support; secondarily it might help with #208.  If it's not wanted I'll maintain it myself or submit it to [wrapdb](https://wrapdb.mesonbuild.com/).


EmielBruijntjes:
I'm happy to merge this if that helps you, but the problem that I have with these kind of PR's is is that I do not use this technology, and I am also not familiar with it. I can therefore not maintain it or can judge what the quality of it is. 

So I wait a little while to see if there are others who think this is helpful too before I press the merge-button.

Issue #261, opened 2018-10-29
Crash when connection lost without close in channel destructor


shanecarey17:
Hello, thanks for the library I have been using it for some time! Recently discovered a possible bug where the state is not set to closed after the connection is lost. This causes an exception to be thrown from the destructor of AMQP::Channel which then calls std::terminate and crashes the program.

Here are the steps in the logic.

0. TCP connection is lost without a proper shutdown from the rabbitmq server
1. AMQP::TcpConnection::process() is called
2. AMQP::TcpConnection::reportError() is called, this closes the socket and sets the file descriptor to -1, then calls handler->onError()
3. In my connection handler onError() I destruct each channel object I had instantiated with the connection object *before* destructing the connection object
4. In the destructor of the channel object there is a check to connected() that the channel is still operating on a valid connection, however the state here is not set correctly and connected() returns true
5. The channel destructor sends a close frame through the connection, this should return false early because the connection is no longer up, but the state in the connection is not set properly either
6. The connection tries to send the buffer through TcpState::send(), which then calls handler->monitor() with the -1 file descriptor
7. The asio handler tries to create a new shared watcher object with that -1 file descriptor, an exception is thrown in _socket.assign()
8. Because this all happened in the destructor of AMQP::Channel, the stack unwinding stops and abort() is called from std::terminate

Is this a proper bug or am I incorrectly destructing the channel object in my handler's onError() before I destruct the connection? It does not look like destructing the connection first would properly set those states in the channel and connection. Thank you

shanecarey17:
This issue is reproducible by disabling (commenting out) heartbeats from the client and connecting to a server (not localhost or docker) then waiting about 2.5 minutes

EmielBruijntjes:
This sounds like a bug indeed. But we have refactored the TCP handling of AMQP-CPP in the last couple of weeks and I do not believe that this scenario can take place any more. Can you verify if this is also true for you?

PR #262, opened 2018-10-30
Close tcp connection in tcpstate after handler onerror

shanecarey17:
This fix prevents a crash described in #261


EmielBruijntjes:
Since we've refactored so much since this report, can you check whether this issue still exists?

PR #268, opened 2018-11-28
Adds a force parameter to Connection::heartbeat()

bmario:
This change allows custom ConnectionHandler to control when the heartbeat is actually sent instead of having to work around the hidden idle state of the ConnectionImpl.

Use case: The ConnectionHandler sets up a timer upon finalizing a send operation, which wakes up after the heartbeat interval and calls `Connection::heartbeat(true)`. 

In the current implementation, the ConnectionImpl would just update the idle state of the connection, but not actually send a heartbeat.

### DIFF
--- a/include/amqpcpp/connection.h
+++ b/include/amqpcpp/connection.h
@@ -96,11 +96,12 @@ public:
 
     /**
      *  Send a ping/heartbeat to the channel to keep it alive
-     *  @return bool
+     *  @param  force       always send the heartbeat, even if the connection is not idle
+     *  @return bool        returns false if the send of the heartbeat failed
      */
-    bool heartbeat()
+    bool heartbeat(bool force=false)
     {
-        return _implementation.heartbeat();
+        return _implementation.heartbeat(force);
     }
 
     /**


EmielBruijntjes:
This pull request could be a workaround for a more serious issue. The whole idle-state was meant to prevent that _unnecessary heartbeats_ are sent. You normally set a timer to send out a heartbeat. I believe that a one-minute interval is normal. A completely idle connection would thus see the following data being sent:

00:01:00 heartbeat
00:02:00 heartbeat
00:03:00 heartbeat

However, if AMQP-CPP knows that other data was also sent over the connection, it skips the heartbeat, because that other data also keeps the connection alive:

00:01:00 heartbeat
00:01:30 some other data
~~00:02:00 heartbeat~~
00:03:00 heartbeat

I believe that the server only disconnects if you are idle for more than two heartbeat intervals, and the third heartbeat is still on time (one-and-half minutes after the previous data) so the second heartbeat could indeed be skipped. However, I do see a an issue when your heartbeat-timer expires a little too late (because your program was busy doing something else!?) or when the heartbeat arrives at the server a too late (network congestion?):

00:01:00 heartbeat
00:01:01 some other data
~~00:02:00 heartbeat~~
00:03:02 heartbeat (2 seconds late because your program was busy)

Now an interval of 2 minutes and 1 seconds is between the data and the next heartbeat, which is too long to keep the connection alive. The way I see it, the whole _idle thing that we implemented could be wrong, and it needs a more fundamental fix.

EmielBruijntjes:
You are right. It looks even worse. I think it is better to forget about this _idle thing, and always send a heartbeat if the connection::heartbeat() is called.

EmielBruijntjes:
I just committed a change that removes this idle-check completely. That make this pull request unnecessary.

Issue #127, opened 2017-04-28
Build Question


mrberman87:
I am having issues compiling strait from the examples in the the README.
I am running this in a VM on VirtualBox.  I am running Ubuntu 16.04.2 LTS with all updates.  I am running AMQP-CPP version 2.7 built from source.

Below is my source code, something simple just to get compilation at this point, but I am getting an undefined reference in TcpConnection.  The compilation output is below the source.

After compiling, I have run ldconfig, and the output of 'ldconfig -p' | grep amqp' shows libamqpcpp.so

`
#include <amqpcpp.h>
#include <iostream>

struct MyTcpHandler : public AMQP::TcpHandler
{
    virtual void onConnected(AMQP::TcpConnection *connection)
    {
        std::cout << "MyTcpHandler: " << __FUNCTION__ << std::endl;
    }

    virtual void onError(AMQP::TcpConnection *connection, const char *message)
    {
        std::cout << "MyTcpHandler: " << __FUNCTION__ << std::endl;
        std::cout << std::string(message) << std::endl;
    }

    virtual void onClosed(AMQP::TcpConnection *connection)
    {
        std::cout << "MyTcpHandler: " << __FUNCTION__ << std::endl;
    }

    virtual void monitor(AMQP::TcpConnection *connection, int fd, int flags)
    {
        std::cout << "MyTcpHandler: " << __FUNCTION__ << std::endl;
    }
};

int main()
{
    std::cout << "Hello from " << __FUNCTION__ << std::endl;

    // create an instance of your own tcp handler
    MyTcpHandler myHandler;

    AMQP::Address address("amqp://localhost/");

    // create a AMQP connection object
    AMQP::TcpConnection connection(&myHandler, address);

    // and create a channel
    //AMQP::TcpChannel channel(&connection);

    // use the channel object to call the AMQP method you like
    //channel.declareExchange("", AMQP::fanout);
    //channel.declareQueue("hello");
    //channel.bindQueue("", "hello", "hello");

    return 0;
}
`

mkdir -p build
g++ -Wall -Wno-deprecated-declarations -g -std=c++14 -pthread -I/usr/include -L/usr/lib -lamqpcpp -Isrc -c src/run.cpp -o build/run.o
g++ -Wall -Wno-deprecated-declarations -g -std=c++14 -pthread -I/usr/include -L/usr/lib -lamqpcpp -Isrc -o run.out build/run.o
build/run.o: In function `main':
[...]

PR #269, opened 2018-11-28
CMakeLists.txt: add_source_directory instead of add_subdirectory

galeone:
This fixes https://github.com/CopernicaMarketingSoftware/AMQP-CPP/issues/127

(at least in my and @alheio cases)

Additional information:

OS: archlinux
build command: 
```
cmake -D AMQP-CPP_LINUX_TCP=ON ..
make
sudo make install
```

Once installed the library, I compile passing the ` -lamqpcp` linker flag.

Without this patch, I got the error  `undefined reference to `AMQP::TcpConnection::~TcpConnection()'`

with the patch I can compile and link the library to my applicatin without any problem

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,9 +51,11 @@ macro (add_sources)
 endmacro()
 
 # add source files
-add_subdirectory(src)
+#add_subdirectory(src)
+aux_source_directory(src src_MAIN)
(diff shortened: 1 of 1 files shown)


aarchetype:
It seems src/CMakeLists.txt and src/linux_tcp/CMakeLists.txt is useless now so I offer you to remove it.

PR #280, opened 2019-01-16
Fix build errors when OpenSSL is installed in non-standard place

koprok:


mvdwerve:
This does not compile if OPENSSL_INCLUDE_DIR was not defined.

Can you define it by default?

The checks failed - for more info go to https://travis-ci.org/CopernicaMarketingSoftware/AMQP-CPP/jobs/480295985

koprok:
@mvdwerve, I saw that it does not compile on the build server, but I still don't understand why. The find_path() call should set OPENSSL_INCLUDE_DIR if there is openssl/ssl.h header on the system. It looks like this header is missing. But then, how does the code compile (before my changes) as it includes this header.

mvdwerve:
No idea, I'm going to check this out locally to see if it works there, and otherwise I'll fix the build file because then there is probably some error there.

koprok:
@mvdwerve, it works locally on my own system and also when building through Vcpkg (https://github.com/koprok/vcpkg/commits/amqpcpp). Actually Vcpkg build is the reason I have added this patch - because amqpcpp has to find and use the OpenSSL headers from Vcpkg.

PR #283, opened 2019-02-04
Check pipe writes in linux_tcp to suppress Wunused-result in GCC

weili-jiang:
Fixing issue when compiling with Wunused-result enabled in GCC.


EmielBruijntjes:
We only throw from constructors. Better return false on failure.

weili-jiang:
Now returning true/false from Pipe::notify and letting TcpResolver record the errno message.

PR #295, opened 2019-05-30
linux_tcp/poll: fix undefined behavior in select

pabigot:
The select(2) system call on Linux now includes the restrict qualifier
on the fd_set pointers passed in, causing a diagnostic in gcc9 for the
call blocking for both read and write.  Pass a pointer to a temporary
copy of the set when two references are required.

Signed-off-by: Peter A. Bigot <pab@pabigot.com>

### DIFF
--- a/src/linux_tcp/poll.h
+++ b/src/linux_tcp/poll.h
@@ -106,8 +106,11 @@ public:
      */
     bool active(bool block)
     {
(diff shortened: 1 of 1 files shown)


pabigot:
Alternatively the `active` method on the `Poll` class could just be removed, since it's never invoked.

PR #298, opened 2019-06-19
fix for incorrect handling of synchronous flag

mvdwerve:
Wireshark after changes

for our test code.


EmielBruijntjes:
Please check if it is possible to just flush the first message from the queue inside the reportSuccess() method instead of all the bookkeeping and storing the old state.

PR #312, opened 2019-12-09
Checked destruction of moved channel

LadMartin:
Added check for _implementation validity in the destructor. After move, the ptr becomes invalid yet close() was still called on it in the destructor. IsValid function was added so the children can also check if the channel was moved or not.


EmielBruijntjes:
Thanks for reporting. This has been fixed in commit 599764e

PR #337, opened 2020-06-26
Fix shared linking issue on Windows

aleserb:
On Windows, a mandatory linking with `ws2_32` is required.


Schoppenglas:
Works fine^^ -- Windows10, 32bit, mingw-8.1.0

PR #353, opened 2020-09-28
Address optional parameters

mvdwerve:
Will allow us to support parameters like `amqp://guest:guest@localhost:5672/vhost?connectTimeout=10&resolveTimeout=20` etc. for parameters `connectTimeout=10` and `resolveTimeout=20` for example.

### DIFF
--- a/include/amqpcpp/address.h
+++ b/include/amqpcpp/address.h
@@ -13,6 +13,11 @@
 #pragma once
 
 /**
+ *  Includes
+ */
+#include <type_traits>
+
+/**
  *  Set up namespace
  */
 namespace AMQP {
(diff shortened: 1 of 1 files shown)


include/amqpcpp/address.h
```diff
@@ -113,8 +123,32 @@ class Address
         // find out where the vhost is set (starts with a slash)
         const char *slash = (const char *)memchr(data, '/', last - data);
 
+        // if there is a slash, we search for the ? for extra options
```
EmielBruijntjes, line 126: Why only if there is a slash?
- amqp://localhost/vhost?param=123
- amqp://localhost/?param=123
- amqp://localhost?param=123

Are all of the above cases covered?

EmielBruijntjes:
Can you also add a method to ask for "const-char-*" options instead of just integers?

PR #359, opened 2020-10-01
use of namespace 

akash246:


EmielBruijntjes:
I prefer explicit use of the std namespace.

PR #360, opened 2020-10-02
Improve docs.

Divyanshu9151:


EmielBruijntjes:
The way I see it is that you are added more errors than you fixed.

PR #372, opened 2020-10-14
Fix merge problem of #345

tarhan:
I've found that within commits starting from July changed `ReceivedFrame` class into `InBuffer`.
So my previous #345 pull request became incompatible.

This short commit fixes problem. Also I've tested it with Python's Celery and it works like with #345.


mvdwerve:
I did not see this PR, but I had fixed it just now! Thanks anyway, closing this PR since it already compiles again.

PR #379, opened 2020-11-13
Add option to select an IP randomly

Aljar:
The default behavior of getaddrinfo is to sort the IPs based on proximity. This request add the option to order the IPs randomly for a connection, so loads are distributed more evenly.

### DIFF
--- a/src/linux_tcp/addressinfo.h
+++ b/src/linux_tcp/addressinfo.h
@@ -11,6 +11,8 @@
  *  Dependencies
  */
 #include <random>
+#include "connectionorder.h"
+#include "address2str.h"
 
 /**
  *  Include guard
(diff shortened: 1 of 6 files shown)


src/linux_tcp/addressinfo.h
```diff
@@ -63,6 +69,25 @@ class AddressInfo
...
+        if (randomOrder)
+        {
+            // We need to seed the random number generator. Normally time is taken
+            // for this. Since we want to have randomness within a second we use
+            // more precision via timeval
+            struct timeval time; 
+            gettimeofday(&time, nullptr);
+
+            // We seed with the precision of miliseconds. 
+            srand((time.tv_sec * 1000) + (time.tv_usec / 1000));
```
rwols, line 86: Isn't it the responsibility of the client of this library to seed the random number generator?

src/linux_tcp/addressinfo.h
```diff
@@ -63,6 +69,25 @@ class AddressInfo
...
+            // for this. Since we want to have randomness within a second we use
+            // more precision via timeval
+            struct timeval time; 
+            gettimeofday(&time, nullptr);
+
+            // We seed with the precision of miliseconds. 
+            srand((time.tv_sec * 1000) + (time.tv_usec / 1000));  
+
+            // shuffle the vector.    
+            std::random_shuffle(_v.begin(), _v.end());
```
rwols, line 89: random_shuffle is removed in C++17. We should use std::shuffle instead.

src/linux_tcp/tcpconnection.cpp
```diff
@@ -23,10 +23,11 @@ namespace AMQP {
  *  Constructor
  *  @param  handler         User implemented handler object
  *  @param  hostname        The address to connect to
+ *  @param  random          Randomly select one of the IP addresses that belong to the hostname
  */
-TcpConnection::TcpConnection(TcpHandler *handler, const Address &address) :
+TcpConnection::TcpConnection(TcpHandler *handler, const Address &address, bool random) :
```
EmielBruijntjes, line 28: Instead of adding endless number of parameters to the TcpConnection, we support parameters inside the address. Better adjust the "address" for this.

rwols (approved):

PR #380, opened 2020-11-16
Fixed the segfault which can happen after address.option("connectionOrder") return nullptr

basanets:
Added check if the connectionOrder receives nullptr and not a valid string. This can be a case if address.option("connectionOrder") returns nullptr.


basanets:
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/commit/21a431d8a83c824e3f37dd5c0cd0c5d5b2aca4df#commitcomment-44233922

EmielBruijntjes:
Great, thanks

PR #388, opened 2021-01-16
Find and Link OpenSSL

theirix:
The patch allows to automatically find and link OpenSSL when `AMQP-CPP_LINUX_TCP` option is enabled.


EmielBruijntjes:
Why do we need this? The AMQP-CPP library should also work on systems where openssl is not installed, especially when no amqps:// addresses are being used.

theirix:
The goal of this patch is to provide include dirs to OpenSSL. That's why we need to locate OpenSSL.

You are right, I should not link to OpenSSL because it is loaded dynamically, fixed it. AMQP-CPP will start even if there is no OpenSSL installed.

PR #390, opened 2021-01-27
Support external authentication

stertingen:
- [x] Introduce `Authentication` base class
- [x] Make `Connection` and `TcpConnection` classes use new `Authentication` class
- [x] Add EXTERNAL `Authentication` class

This PR aims to replace #93


stertingen:
@EmielBruijntjes Would you please take a look at this?

stertingen:
If desired I could change the PR to use less `std::shared_ptr`. The conntection object would then store authentication mechanism and response instead of a shared pointer to the authentication object. More simple but maybe less flexible.

EmielBruijntjes:
I indeed do not like the shared-pointer mechanism, especially not that it is exposed to "user space", that implementation-specific detail (that it uses a shared pointer) should be encapsulated inside AMQP-CPP.

stertingen:
> I indeed do not like the shared-pointer mechanism, especially not that it is exposed to "user space", that implementation-specific detail (that it uses a shared pointer) should be encapsulated inside AMQP-CPP.

Okay, the PR is updated to use no `std::shared_ptr`.

plopp:
It seems this is failing in ci because of docker pull rate limiting?

```
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
The command "docker run -e "DEBIAN_FRONTEND=noninteractive" -d --name ubuntu-test-container -v $(pwd):/travis ubuntu:focal tail -f /dev/null" failed and exited with 125 during .
```

plopp:
@stertingen would you mind retriggering the CI?

stertingen:
> @stertingen would you mind retriggering the CI?

How would I do this?

plopp:
> > @stertingen would you mind retriggering the CI?
> 
> How would I do this?

Try closing the PR and reopening it. If that does not work, we would need help from someone with write access to the repo to retrigger it. Of course you could just push a meningless commit to your fork I guess.

The reason I want you to try to get CI to rerun is I think the build failed because of rate limiting on the docker hub. Nothing to do with your change, becasue your code looks fine.

As long as it does not build properly it won't get traction for a merge. Thanks!

plopp:
Great @stertingen! Let's hope @EmielBruijntjes can decide on a "go or no-go" now that it passes the checks!

trevorperrin:
One issue I see with this change is that Address already includes an Authentication object (Login), which means that two authentication objects are created when creating a TcpConnection.  

> AMQP::Login myLogin("", "");
> AMQP::Address address(host, port, myLogin, vhost, true);
> AMQP::TcpConnection connection(&myHandler, address, AMQP::ExternalAuth());

Could Address be modified to use Authentication instead, in which case perhaps TcpConnection doesn't need the third parameter?

This is one reason why my initial inclination was to modify Login to provide external authentication like pull request #93, and my pull request #458 (which is an alternate but very similar approach).  Either way works, it's a question of which is more desirable

EmielBruijntjes:
I agree with @trevorperrin. The structure of the code and how the Address, Login and Authentication classes interact needs a bit more attention.

PR #397, opened 2021-06-07
Properly unmark synchronized channel mode

### LINKED ISSUE #396: Multiple synchronous operations active at a time
The documentation recommends that
> ... you should wait for the onReady() callback to be called before you send any other instructions over the channel.

and the referred example also shows that instructions are sent from the `onReady()` callback.

This leads, however, to multiple synchronous operations being active at a time. This is not a problem as long as the server responds in-order, but the server may (and does) respond out-of-order on occasion. In such a case, the response is not dispatched to the proper callback (and, depending on the instruction, the library will cease to work).

This is caused by the fact that `flush()` is called after the `onReady()` callback of the channel, and `flush()` resets the `_synchronous` state of the channel:
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/dc9d88f4b1bad73a409f89e4ed688c653eaea229/include/amqpcpp/channelimpl.h#L629-L633
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/dc9d88f4b1bad73a409f89e4ed688c653eaea229/src/channelimpl.cpp#L825-L828
   
If the `onReady()` callback issued at least two instructions, both of them become active. The first one is sent to the server immediately and the channel is properly put into `_synchronous` state, the second one (and any subsequent) are properly queued. After the callback has finished execution, `flush()` is called, which will reset the `_synchronous` flag to `false` and therefore causes the second instruction to be pop'ed from `_queue` and sent to the server before any other response from the server has been received.

lukasgeyer:
Unmarking the channel from synchronous operation in flush() might cause
multiple active synchronous operations, as flush() may be called
multiple times for a received frame with the possibility of other
frames being queued in between.

Fixes #396.

### DIFF
--- a/include/amqpcpp/channelimpl.h
+++ b/include/amqpcpp/channelimpl.h
@@ -623,11 +623,11 @@ public:
         // if we are still in connected state we are now ready
         if (_state == state_connected) _state = state_ready;
         
-        // the last (possibly synchronous) operation was received, so we're no longer in synchronous mode
-        if (_synchronous && _queue.empty()) _synchronous = false;
-
         // inform handler
         if (_readyCallback) _readyCallback();
+
+        // the last (possibly synchronous) operation was received, so we're no longer in synchronous mode
+        if (_synchronous && _queue.empty()) _synchronous = false;
 
         // if the monitor is still valid, we flush any waiting operations 
         if (monitor.valid()) flush();


EmielBruijntjes:
The flush() method is called in other cases too. Have you checked whether those calls are still ok?

lukasgeyer:
There is indeed another relevant invocation:
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/dc9d88f4b1bad73a409f89e4ed688c653eaea229/src/deferredget.cpp#L62-L69
Fixing it there would mean to extend the `ChannelImpl` API though, as there is no public setter to the `_synchronous` property.

Maybe it is better to revert the current change and reverse the invocation of the `onReady()` callback, so that in fact all instructions invoked there are queued (because `_synchronous` may still be set `true` then and `flush()` will properly reset it):
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/dc9d88f4b1bad73a409f89e4ed688c653eaea229/include/amqpcpp/channelimpl.h#L626-L630

@EmielBruijntjes what do you think?

lukasgeyer:
@EmielBruijntjes are you fine with the updated fix or do you need any additional information?

EmielBruijntjes:
Sorry that my response was not earlier. Although I think that this fix is logically correct, I still am not sure if I like it. With this change we will now buffer all outgoing frames even when that is not needed. From 'user space' you can either take two approaches when using AMQP-CPP:

1. You let AMQP-CPP take care of buffering for frames that should be delayed until an earlier instructions are ready.
2. Or, if you don't want to pay the penalty for buffering, you implement onReady() callbacks and send the instructions yourself when you know for sure that no buffering will take place.

But with this latest proposal, it seems that users who try to avoid buffering (and go for option 2), will end up having buffers anyway.

EmielBruijntjes:
Can you take a look at #398? Does that solve the issue too?

lukasgeyer:
@EmielBruijntjes I see.

#398 seems to work quite fine, if multiple instructions are issued in the `onReady()` callback the first one is sent immediately and the other ones are queued and then sent one after another if a response for the previous one has been received.

Thanks for the fix, feel free to reject this one then.

EmielBruijntjes:
Also fixed in #398

PR #400, opened 2021-07-01
Prefer make_shared over calling std::shared_ptr constructor

rwols:
Noticed this while investigating project 35447 (or see 40168). The std::shared_ptr constructor, when given a raw pointer, is forced to allocate its control block separately. When using std::make_shared, it is able to allocate sizeof(control block) + sizeof(T) in one allocation.

### DIFF
--- a/src/array.cpp
+++ b/src/array.cpp
@@ -32,7 +32,7 @@ Array::Array(InBuffer &frame)
         charsToRead -= (uint32_t)field->size();
 
         // add the additional field
-        _fields.push_back(move(field));
+        _fields.push_back(std::move(field));
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/array.cpp
```diff
@@ -25,14 +25,14 @@ Array::Array(InBuffer &frame)
...
-        Field *field = Field::decode(frame);
+        auto field = Field::decode(frame);
         if (!field) continue;
 
         // less bytes to read
         charsToRead -= (uint32_t)field->size();
 
         // add the additional field
-        _fields.push_back(std::shared_ptr<Field>(field));
+        _fields.push_back(move(field));
```
EmielBruijntjes, line 35: should this not be std::move()?

EmielBruijntjes (changes requested):

src/array.cpp
rwols: Not necessarily due to ADL :) depends on what the code-style wants I suppose. I've added the std:: prefix.

PR #409, opened 2021-07-28
Three small improvements

Niederb:
Feel free to pick the changes you like

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ would be:
 ```bash
 mkdir build
 cd build
-cmake .. [-DAMQP-CPP_BUILD_SHARED] [-DAMQP-CPP_LINUX_TCP]
+cmake .. [-DAMQP-CPP_BUILD_SHARED=ON] [-DAMQP-CPP_LINUX_TCP=ON]
 cmake --build . --target install
 ```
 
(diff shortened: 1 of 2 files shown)


include/amqpcpp/monitor.h
```diff
@@ -54,7 +54,7 @@ class Monitor
     Monitor(Watchable *watchable) : _watchable(watchable)
     {
         // register with the watchable
-        _watchable->add(this);
+        if (_watchable) _watchable->add(this);
```
EmielBruijntjes, line 57: Did you run into an issue because of this? Afaik we never call Monitor(nullptr).

EmielBruijntjes (changes requested):
Could you please explain why the change in Monitor was made? Maybe it is caused by an issue somewhere else.

include/amqpcpp/monitor.h
Niederb: I didn't run into actual issues. Just when looking at the code I saw that _watchable is allowed to be nullptr in all other methods/constructors and thought it is more consistent to also handle it gracefully in this constructor.

EmielBruijntjes: No it is not an issue, in the constructor it can never be null (in the other methods it can because it could be set to null _after_ construction). Can you undo this?

Niederb: Ok, I removed the corresponding commit

rwols (approved):

PR #412, opened 2021-07-28
Avoid MSVC compiler warnings in Envelope

### LINKED ISSUE #399: warnings in amqp-cpp that are treated as errors in our build
hello.

we are using AMQP-CPP library in our project and having troubles building it, because several headers have warnings in them, and they are treated as errors:

at
```
envelope.h line 68
envelope.h line 125
libevent.h line 67
```

```
C:\...\.conan\data\amqp-cpp\4.3.11\_\_\package\8cf01e2f50fcd6b63525e70584df0326550364e1\include\amqpcpp/envelope.h(68): warning C4244: 'argument': conversion from 'uint64_t' to 'uint32_t', possible loss of data 
...\include\amqpcpp/envelope.h(125): warning C4244: 'argument': conversion from 'const uint64_t' to 'uint32_t', possible loss of data (compiling source file ...) 
...\include\amqpcpp/libevent.h(67): warning C4244: 'argument': conversion from 'intptr_t' to 'int', possible loss of data (compiling source file ..) 
```

----
we believe it is a good practice to treat warnings as errors.
could you please fix those?

rwols:
Fixes #399

### DIFF
--- a/include/amqpcpp/envelope.h
+++ b/include/amqpcpp/envelope.h
@@ -64,8 +64,11 @@ public:
     Envelope(InBuffer &buffer) : MetaData(buffer)
     {
         // extract the properties
+        // @todo: is this correct? can body sizes be uint64?
         _bodySize = buffer.nextUint64();
-        _body = buffer.nextData(_bodySize);
+
+        // @todo: is casting to uint32 correct?
+        _body = buffer.nextData(static_cast<uint32_t>(_bodySize));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


SebDieBln:
I just ran into the same warnings. Is there any progress with this PR?

EmielBruijntjes:
This pull request is not correct, as it does not solve the actual problem. The casting to 32 bit should not be necessary, as bigger payloads are valid in the AMQP protocol.

SebDieBln:
> This pull request is not correct, as it does not solve the actual problem. The casting to 32 bit should not be necessary, as bigger payloads are valid in the AMQP protocol.

I agree. I just looked at `Message::setBodySize()` and it contains an explicit check for such messages. So I guess the actual problem is in `InBuffer::nextData()` which takes a `uint32_t` as its `size` parameter instead of a `size_t`.

SebDieBln:
@EmielBruijntjes Thanks for fixing the warnings in 328d9d12de10475b4e389d0f54ca054867a3af24!

Issue #420, opened 2021-08-15
sigsegv when server is inactive for a long time (use after free)


gusev-vitaliy:
**Describe the bug**

If server doesn't reply on the heartbeats,  AMQP::LibEvHandler::Wrapper::onExpired() closes connection without stopping repeatable timer. Destructor ~Wrapper() doesn't stop it either. All above leads to use-after-free and to SEGSEGV or accerts.

Stack:

```text
(lldb) bt
* thread #1, name = 'test', stop reason = signal SIGSEGV
  * frame #0: 0x0000000800378e4e libev.so.4`ev_feed_event(loop=0x00000008003827b8, w=0x0000000801027028, revents=256) at ev.c:2319:5
    frame #1: 0x000000080037ac31 libev.so.4`ev_run [inlined] feed_reverse_done(loop=0x00000008003827b8, revents=256) at ev.c:2342:5
    frame #2: 0x000000080037abf4 libev.so.4`ev_run [inlined] timers_reify(loop=0x00000008003827b8) at ev.c:3836
    frame #3: 0x000000080037a98d libev.so.4`ev_run(loop=0x00000008003827b8, flags=0) at ev.c:4174
    frame #4: 0x00000000002072ff test`main at test.cpp:16:2
    frame #5: 0x0000000000207090 test`_start(ap=<unavailable>, cleanup=<unavailable>) at crt1_c.c:75:7
```

**Expected behavior and actual behavior**

Wrapper::callback() is not called after ~Wrapper() and _timer is not active.

**Sample code**
```c++
#include <ev.h>
#include <amqpcpp.h>
#include <amqpcpp/libev.h>
                                                                                                                                                                          
int main()                                                                                                                                                                
{                                                                                                                                                                         
    auto *loop = EV_DEFAULT;                                                                                                                                              
                                                                                                                                                                          
    // make a connection                                                                                                                                                  
    AMQP::LibEvHandler handler(loop);                                                                                                                                     
    AMQP::Address address("amqp://guest:guest@localhost/");                                                                                                               
    AMQP::TcpConnection connection(&handler, address);                                                                                                                    
    AMQP::TcpChannel channel(&connection);                                                                                                                                
                                                                                                                                                                          
    // run the loop                                                                                                                                                       
    ev_run(loop, 0);                                                                                                                                                      
    return 0;
}
```
**How to reproduce**

****Method 1****
[...]

PR #421, opened 2021-08-15
Stop timer at Wrapper::onExpired() when server is inactive (#420)

gusev-vitaliy:
Note, that this uses close(true) - do not wait for a lot of time since server hasn't responded  for about 1.5 timeout seconds.

- [ ] check ev_ref/ev_unref

### DIFF
--- a/include/amqpcpp/libev.h
+++ b/include/amqpcpp/libev.h
@@ -254,9 +254,11 @@ private:
             {
                 // the server was inactive for a too long period of time, reset state
                 _next = _expire = 0.0; _timeout = 0;
-                
+
+                ev_timer_stop(_loop, &_timer);
+
                 // close the connection because server was inactive
-                return (void)_connection->close();
+                return (void)_connection->close(true);
             }
             else if (now >= _next)
             {


include/amqpcpp/libev.h
```diff
@@ -254,9 +254,11 @@ class LibEvHandler : public TcpHandler
             {
                 // the server was inactive for a too long period of time, reset state
                 _next = _expire = 0.0; _timeout = 0;
-                
+
+                ev_timer_stop(_loop, &_timer);
+
```
EmielBruijntjes, line 259: I don't get this change. The fact that the timer expires implicitly means that it is stopped, so there shouldn't be a need to explicitly stop it, right?

gusev-vitaliy: Setting _timer.repeat means that timer becomes repeating. Look at man ev(3):

>    "ev_timer" - relative and optionally repeating timeouts
>            Timer watchers are simple relative timers that generate an event after a given time, and optionally repeating in regular intervals after that.

Callback of repeating  timer is called regularly until ev_timer_stop() is called.

Please look at the test code.

gusev-vitaliy: @EmielBruijntjes To see repeat behaviour you can run this simple test:
```c++
#include <iostream>
#include <ev.h>

static void
callback(struct ev_loop *loop, struct ev_timer *timer, int revents)
{
    std::cout << "callback" << std::endl;
    return;
}

int main()
{
    ev_timer timer;
    struct ev_loop *loop = EV_DEFAULT;

    ev_init(&timer, callback);
    timer.repeat = 3;
    ev_timer_start(loop, &timer);
    ev_run(loop, 0);

    return 0;
}
```

EmielBruijntjes: Yes you are right, I forgot about that repeat-setting, I will check it out a bit later

gusev-vitaliy: I think better fix is to call ev_timer_stop() unconditionally in ~Wrapper(), i.e. do not check timed(). Man says:

>  "ev_TYPE_stop" (loop, ev_TYPE *watcher)
         ...
         If you want to free or reuse the memory used by the watcher it
         is therefore a good idea to always call its "ev_TYPE_stop" function.

However it seems the fix should contain ev_ref/ev_unref calls as well. Could you point what should be achieved in the end by using those? I see ev_timer_again() call is followed by ev_unref() in onExpired(). But ev_timer_again() call in Wrapper::start() is not followed. So I am confused.

EmielBruijntjes: Sorry for being a little late. I made a mistake with setting the repeat property and the call to 'ev_timer_again()'. I did not realize that this would create a _repeating_ timer. Our code is written with the idea in mind that the timer expires _once_, and then is re-armed for the _next_ timerout. I thought that the ev_timer_again() function would only trigger the timer one more time. So this code was wrong:
```
// reset the timer to trigger again later
_timer.repeat = std::min(_next, _expire) - now;

// restart the timer
ev_timer_again(_loop, &_timer);
```
Should probably be replaced with this to get it to function the way I intended:
```
// reset the timer to trigger again later
ev_timer_set(&_timer, std::min(_next, _expire) - now, 0.0);

// restart the timer
ev_timer_start(_loop, &_timer);
```

EmielBruijntjes: Could you be so helpful to adjust your pull request and see if this change fixes it too?

happykeyboard: I observe a similar segfault, on tagged version v4.1.6. I get a segfault on while attempting to recover from a disconnect event. To recover from a disconnect event, I instantiate a new `AMQP::TcpConnection`, and pass it to a new `AMQP::TcpChannel` object. Segfault follows shortly with a backtrace that looks like this: 
```
0: new_allocator<AMQP::Monitor*>::construct<AMQP::Monitor*, AMQP::Monitor* const&> at new_allocator.h:136
1: std::allocator_traits<std::allocator<AMQP::Monitor*> >::construct<AMQP::Monitor*, AMQP::Monitor* const&> at alloc_traits.h:475
2: std::vector<AMQP::Monitor*, std::allocator<AMQP::Monitor*> >::push_back ( at stl_vector.h:943
3:  AMQP::Watchable::add at AMQP-CPP/include/amqpcpp/watchable.h:50
4: AMQP::Monitor::Monitor at AMQP-CPP/include/amqpcpp/monitor.h:57
5: AMQP::TcpConnection::close at AMQP-CPP/src/linux_tcp/tcpconnection.cpp:115
6: AMQP::LibEvHandler::Wrapper::onExpired at AMQP-CPP/include/amqpcpp/libev.h:251
7: AMQP::LibEvHandler::Wrapper::callback at AMQP-CPP/include/amqpcpp/libev.h:214
8: ev_invoke_pending ()
9: ev_run ()
```

per @EmielBruijntjes  suggestion, the following diff indeed fixes the segfault
```
diff --git a/include/amqpcpp/libev.h b/include/amqpcpp/libev.h
index 110f251..0f61725 100644
--- a/include/amqpcpp/libev.h
+++ b/include/amqpcpp/libev.h
@@ -267,12 +267,12 @@ private:
                 // sent only after _timout/2 seconds again _from now_ (no catching up)
                 _next = now + std::max(_timeout / 2, 1);
             }
-            
+
             // reset the timer to trigger again later
-            _timer.repeat = std::min(_next, _expire) - now;
-            
+            ev_timer_set(&_timer, std::min(_next, _expire) - now, 0.0);
+
             // restart the timer
-            ev_timer_again(_loop, &_timer);
+            ev_timer_start(_loop, &_timer);
             
             // and because the timer is running again, we restore the refcounter
             ev_unref(_loop);
```
similar fix should be probably applied to `uint16_t start(uint16_t timeout)` method, but testing suggests that reconnect times increase.

When can we expect this bug to be fixed?
Thanks!

gusev-vitaliy:
I think the issue #420  is closed by #423, so we can close this PR.

include/amqpcpp/libev.h
gusev-vitaliy: > When can we expect this bug to be fixed?
> Thanks!

It must be fixed in #423

PR #431, opened 2021-09-24
Use GNUInstallDirs cmake module

matwey:
Some Linux distributions prefer to use /usr/lib64 destination for libraries at
64bit architectures. To be as generic as possible, use GNUInstallDirs module to
resolve correct destination paths.

In case, the paths can still be overridden from the command line as the following:

```
cmake .. -DCMAKE_INSTALL_LIBDIR:PATH=lib
```


theirix:
Resolved by similar PR #515

PR #437, opened 2021-11-23
add x-message-deduplication exchange type

TonyRoussel:
This exchange type exist on RabbitMQ server with the plugin rabbitmq_message_deduplication
This plugin code is available here: https://github.com/noxdafox/rabbitmq-message-deduplication

### DIFF
--- a/src/channelimpl.cpp
+++ b/src/channelimpl.cpp
@@ -290,6 +290,7 @@ Deferred &ChannelImpl::declareExchange(const std::string &name, ExchangeType typ
     else if (type == ExchangeType::topic)           exchangeType = "topic";
     else if (type == ExchangeType::headers)         exchangeType = "headers";
     else if (type == ExchangeType::consistent_hash) exchangeType = "x-consistent-hash";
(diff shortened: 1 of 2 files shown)


src/channelimpl.cpp
```diff
@@ -290,6 +290,7 @@ Deferred &ChannelImpl::declareExchange(const std::string &name, ExchangeType typ
     else if (type == ExchangeType::topic)           exchangeType = "topic";
     else if (type == ExchangeType::headers)         exchangeType = "headers";
     else if (type == ExchangeType::consistent_hash) exchangeType = "x-consistent-hash";
+    else if (type == ExchangeType::consistent_hash) exchangeType = "x-message-deduplication";
```
rwols, line 293: this should be `message_deduplication` instead of `consistent_hash`?

TonyRoussel: yes indeed really sorry I fix right now

TonyRoussel: I [sent a PR](https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/438) with the fix :/

PR #456, opened 2022-03-21
Allow user to pass a pre-configured SSL_CTX to the client.

pgit:
When using a secure connection, the `onSecuring()` callback is invoked with a pointer to a freshly created OpenSSL `SSL` object. You can then install certificates etc. on this session object.

This session object is created from a 'template' `SSL_CTX` object, which stores the defaults for all newly created SSL sessions. Some applications might have a pre-configured `SSL_CTX` and use that for configuring the SSL connection.

This patch adds a way to pass a pre-configured `SSL_CTX` to the connection, which is used instead of the one that is constructed internally.

### DIFF
--- a/include/amqpcpp/linux_tcp/tcpconnection.h
+++ b/include/amqpcpp/linux_tcp/tcpconnection.h
@@ -128,6 +128,17 @@ private:
     }
     
     /**
+     *  Method that is called to get a user-provided SSL context
+     *  @param  state
+     *  @return SSL_CTX*
+     */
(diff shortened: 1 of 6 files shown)


EmielBruijntjes:
Have you also tried to add extra methods to the TcpHandler class for initialization/construction of the ssl context? I find the extra parameter a bit odd.

pgit:
Added a new overload of onSecuring() that userspace can use to return a SSL_CTX. Not quite sure about naming here. Also, are there no concerns about ABI breakage from adding a new method to the vtable?

PR #458, opened 2022-04-18
Allow user to provide an alternative Authentication mechanism

trevorperrin:
The modifications to login.h and connectionstartframe.h allow a user to optionally provide an authentication mechanism other than the default PLAIN mechanism (username and password).

### DIFF
--- a/include/amqpcpp/login.h
+++ b/include/amqpcpp/login.h
@@ -78,7 +78,7 @@ public:
     }
 
     /**
-     *  Constructor
+     *  Constructor with PLAIN authentication mechanism
      *  @param  user
      *  @param  password
      */
--- a/include/amqpcpp/login.h
+++ b/include/amqpcpp/login.h
@@ -39,20 +39,43 @@ private:
      */
     std::string _password;
 
+    /**
+     *  The security mechanism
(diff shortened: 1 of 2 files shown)


include/amqpcpp/login.h
```diff
@@ -39,28 +39,79 @@ class Login
...
+
+    /**
+     *  Constructor with custom authentication mechanism
+     *  @param  user
+     *  @param  password
+     *  @param  mechanism
+     *  @param  response
+     */
+    Login(const char *user, const char *password, const char *mechanism, const char *response)
+        : _user(user), _password(password), _mechanism(mechanism), _response(response) {}
```
EmielBruijntjes, line 114: Can you add more in-code documentation? I don't understand the use case where it is up to the caller to supply a username, password, mechanism and response. What sort of values is one supposed to pass here?
```c++
Login mylogin("my-username", "my-password", "my-special-login-mechanism", "bla bla bla");
```
Is this permitted?

EmielBruijntjes (changes requested):
I don't understand the change, and we need (at least) more in-code documentation to explain to users why they have to pass a username, password, mechanism and response.

PR #461, opened 2022-04-28
Allow passing pointers to arbitrary types to Channel::publish()

SebDieBln:
This PR allows to write code like this
```
void foo(AMQP::Channel & ch, someType & data)
{
    ch.publish("", "", &data, sizeof(data));
}
```
whereas before you had to explicitly cast like this
```
void foo(AMQP::Channel & ch, someType & data)
{
    ch.publish("", "", reinterpret_cast<char*>(&data), sizeof(data));
}
```

### DIFF
--- a/include/amqpcpp/channel.h
+++ b/include/amqpcpp/channel.h
@@ -405,7 +405,7 @@ public:
      */
     bool publish(const std::string &exchange, const std::string &routingKey, const Envelope &envelope, int flags = 0) { return _implementation->publish(exchange, routingKey, envelope, flags); }
     bool publish(const std::string &exchange, const std::string &routingKey, const std::string &message, int flags = 0) { return _implementation->publish(exchange, routingKey, Envelope(message.data(), message.size()), flags); }
-    bool publish(const std::string &exchange, const std::string &routingKey, const char *message, size_t size, int flags = 0) { return _implementation->publish(exchange, routingKey, Envelope(message, size), flags); }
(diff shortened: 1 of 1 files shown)


SebDieBln:
If it is desired I could extend this PR by modifying `AMQP::Envelope` to also accept pointers to arbitrary data types.

EmielBruijntjes:
Hmm, not sure whether I see the necessity for this change. But if we do decide to change the payload type from 'char *' to 'void *', we probably also want to change this in other places so that the library stays consistent. An example would be the Envelope-constructor and the Envelope::body() methods. Otherwise messages are published as 'void *', but consumed as 'char *'.

SebDieBln:
> Check if you can keep the type of the payload consistent and use the same type (like void*) everywhere where payload data is passed or returned.

Thanks for the feedback. I agree and I also prefer consistency throughout the library. I will come up with the necessary changes soon.

> not sure whether I see the necessity for this change.

I understand. People could be tempted to pass non-POD types and that would surely not work. I will still come up with the requested changes and look around how other libraries handle similar situations. Then we can decide to merge or not.

SebDieBln:
I close the PR because I myself don't see the need for this change anymore. Accepting any pointer type would make it easy to use the library in the wrong way (passing non-POD types).

I found other libraries (i.e. Qt's [ByteArray](https://doc.qt.io/qt-6/qbytearray.html)) that also only accept `char*` and not `void*` when working with buffers.

PR #463, opened 2022-05-31
libboostasio: close connection on timeout

ricardocrl:
The implementation is inspired in the approach used in libev.h. It will now handle `timeout` from `onNegotiate()` in the same manner:
- heartbeat every: timeout / 2
- close connection after: timeout * 1.5

As opposed to libev.h I opted for 2 separate timer (sending heartbeat + expire connection), for readability and maintainability purposes.

This PR addresses issue:

### DIFF
--- a/include/amqpcpp/libboostasio.h
+++ b/include/amqpcpp/libboostasio.h
@@ -1,11 +1,11 @@
 /**
  *  LibBoostAsio.h
  *
- *  Implementation for the AMQP::TcpHandler for boost::asio. You can use this class 
- *  instead of a AMQP::TcpHandler class, just pass the boost asio service to the 
+ *  Implementation for the AMQP::TcpHandler for boost::asio. You can use this class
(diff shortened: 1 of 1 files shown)


include/amqpcpp/libboostasio.h
```diff
@@ -468,24 +513,24 @@ class LibBoostAsioHandler : public virtual TcpHandler
...
         if (iter == _watchers.end()) return 0;
 
         // set the timer
-        iter->second->set_timer(connection, interval);
+        iter->second->set_timers(connection, timeout);
 
-        // we agree with the interval
-        return interval;
+        // we agree with the timeout
+        return timeout;
```
ricardocrl, line 533: This renaming is to achieve consistency with libev.h.

Although "interval" is the most used term in the library, I feel that calling "interval" wouldn't be intuitive to read "heartbeat interval = interval / 2". So I agree more with the "timeout" term chosen in libev.h

PR #496, opened 2023-03-10
Add version header with AMQPCPP_VERSION preprocessor symbol

Chardrazle:
A bit of cmake machinery that stamps out a "version.h" file based on the existing values in _CMakeLists.txt_.
The _version_template.h_ file is used to construct the version header.  The initial "version.h" file is created in the build (binary) dir so as not to pollute the source tree.  It is installed into ".../include/amqpcpp/".


EmielBruijntjes:
What would the recommended workflow be when we bring out a new version? Currently, after I brought out a new version, I already (manually) update the version numbers in Makefile and CMakeLists.txt to prepare for the _next_ version.

Chardrazle:
Does anything change?
Presumably at the point where you’ve decided the release is ready, that’s the snapshot point where the values in the make files truly represent that version (as opposed to _working towards_).
You then set the tag to mark it.  After that you update the values for the next targeted version.
If I were to download, say, v4.3.20, the `VERSION_XYZ` values in its _CMakeLists.txt_ file specify 4, 3 and 20.  So a cmake, make, make install will always be bound to those values.

PR #510, opened 2023-08-24
implemented heartbeat timer for LibUV example

thidas1290:
broker disconnects the client when heartbeats are not sent in the given LibUV example

rabbitmq-timeout

- implemented a timer for sending heartbeats.

### DIFF
--- a/include/amqpcpp/libuv.h
+++ b/include/amqpcpp/libuv.h
@@ -167,11 +167,29 @@ private:
     uv_loop_t *_loop;
 
     /**
+     *  timer
+     *  @var uv_timer_t*
+     */
+    uv_timer_t *_timer;
+
+    /**
      *  All I/O watchers that are active, indexed by their filedescriptor
      *  @var std::map<int,Watcher>
      */
     std::map<int,std::unique_ptr<Watcher>> _watchers;
 
+    /**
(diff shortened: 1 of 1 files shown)


EmielBruijntjes:
For LibEv this has already been implemented, but the code is much more complex over there. For example:

- It is possible for a single LibEv instance to monitor multiple connections (and thus have multiple timers active)
- What if the object is destructed before any heartbeat timers were activated?
- We also take into account that no heartbeats have to be sent if we already sent some other kind of data
- We also close the connection if the server failed to send a heartbeat in time
- We do some stuff with interval/2 to ensure that we're really on time with sending the heartbeat

Can you take a look at such things? Otherwise it might be better to just disable heartbeats.

EmielBruijntjes (changes requested):
Take a look at the LibEv implementation, which has much more details to send and process heartbeats. Is it possible to turn that class into a libuv implementation?

thidas1290:
@EmielBruijntjes updated the LibUV heartbeat implementation referencing LibEv example such that,

- can have multiple connections, each connection having a heartbeat timer and multiple fd watchers.
- overall heartbeat sending logic remains untouched, changes are made only to time conversions and LibUv specific functions

Issue #424, opened 2021-09-06
Boost handler doesn't work with SSL


svyatogor:
**Describe the bug**
Current implementation of Boost Asio handler doesn't play well with SSL and causes very fast memory leak with 100% cpu usage. The underlying reason is that unlike uv and ev the implementations it doesn't like `monitor` being called more than once with the same `fd` and `flags`, which just results in multiple handlers being created. This doesn't seem to affect non-ssl connections but as SSL will "repeat" on SSL_ERROR_WANT_READ it immediately becomes a problem.

I am well away that this implementation is not done by the library author, but I still decided to leave this bug report here in cases someone else faces the same issue.

In case someone finds it useful I am attaching my implementation of the handler, which in addition supports heartbeats. It was tested with C++ 17 and I am not certain if it will work on lower versions out of the box. 

https://gist.github.com/svyatogor/6356dc1b4bea0033c086f7be2bd77843

JCacho:
Thank you for your help as we have spent some time facing the same issue.

paolopas:
Thanks for the example, I understood the problem of repeated calls to `LibBoostAsioHandler::monitor` with the same parameters (`fd` and `flags`), exactly as you say, spurious callbacks are created.

I'm refactorying `libboostasio` (see #550) and will take your report into account, luckily the change I have to make is minimal.

paolopas:
In reality, the fix is ​​not as simple as adjusting a condition...

Issue #238, opened 2018-07-09
AMQP::LibBoostAsioHandler::Watcher "memory leak"


msukalski:
The Watcher code uses 2 boolean flags, _read_pending and _write_pending, to attempt to prevent more than one read and write ASIO operation from being posted at a time. Unfortunately, the existing implementation is not thread-safe, since changes in the boolean values are not necessarily visible to all threads at the same time. The net result is that the ASIO queues end up with requests that are never processed, and it appears like a slow but always growing memory leak that will eventually consume all available memory.

I've attached my changes to these flags (std::atomic<bool> ...) that require C++11 support. compare_exchange_strong() is used to prevent more than one thread from posting a read or write operation at a time. This version of libboostasio.h has also been modified for using asio in standalone mode (and all other boost support has been removed through the use of lambdas everywhere). This code has been tested and verified. There are other ways of addressing this issue, of course, so your mileage may vary...

[libboostasio.h.txt](https://github.com/CopernicaMarketingSoftware/AMQP-CPP/files/2176870/libboostasio.h.txt)

EHadoux:
Is it related to #237? Because after closing the connection watchers are all that is left.

zerodefect:
Can I confirm that once you introduced atomic read/writes on the pending variables that the memory leak(s) (you are reporting) ceased?

mihacooper:
I have found the similar problem with `AMQP::LibBoostAsioHandler`. This is my valgrind's `massif` output
![asio_memory_leak](https://user-images.githubusercontent.com/5420190/63760002-28332b80-c8c7-11e9-84df-01a800bf2166.png)

Patch proposed by @msukalski looks like fix the problem. 
![asio_memory_fix](https://user-images.githubusercontent.com/5420190/63760221-937cfd80-c8c7-11e9-8d20-e73408c28464.png)

I have made a real patch from the posted code:
[libboostasio.h.patch.txt](https://github.com/CopernicaMarketingSoftware/AMQP-CPP/files/3545399/libboostasio.h.patch.txt)

But I have nothing to say about its correctness and workability.

zerodefect:
Ok. Good to know. I'll look at this evening.  Thanks for your investigative work.

mvdwerve:
[...]

PR #534, opened 2024-08-19
Reimplement libboostasio

matwey:
Provide better implementation for Boost.ASIO library support. New implementation have the following advantages over existing one:
 * all OS supported
 * TLS/SSL is supported natively on all OS using Boost.ASIO
 * Boost.ASIO multithreading support is safe and correct (see examples/libboostasio_multithreading.cpp)


corporategoth:
FYI, I just tested this

You need to change line 120-122 from:

```
		for (const auto& buffer: _rx_streambuf.data()) {
			consumed_total += _connection.parse(reinterpret_cast<const char*>(buffer.data()), buffer.size());
		}
```

to:
```
		auto const_buffers = _rx_streambuf.data();
		for (const auto* buffer = boost::asio::buffer_sequence_begin(const_buffers);
				buffer != boost::asio::buffer_sequence_end(const_buffers); ++buffer) {
			consumed_total += _connection.parse(reinterpret_cast<const char*>(buffer->data()), buffer->size());
		}
```

The current code as written does not compile with at least boost 1.87.0

Otherwise, the minimal testing I've done seems to show this patch works, and allows me to connect with SSL appropriately, and even, because I am doing the connection myself, I can adjust certificate verification and support mTLS (which RabbitMQ supports).

@matwey could you update this as per the above, and @EmielBruijntjes , could we get this merged?
It breaks anyone currently using libboostasio.h - but this IS required to both better integrate boost::asio, and to ensure SSL works correctly without reported memory leaks, etc (see https://github.com/CopernicaMarketingSoftware/AMQP-CPP/issues/424, https://github.com/CopernicaMarketingSoftware/AMQP-CPP/issues/238, https://github.com/CopernicaMarketingSoftware/AMQP-CPP/issues/207)

matwey:
@corporategoth I need some time to return into the context. Don't expect me to reply soon.

matwey:
Your code won't work as you expected:

```c++
		auto const_buffers = _rx_streambuf.data();
		for (const auto* buffer = boost::asio::buffer_sequence_begin(const_buffers);
				buffer != boost::asio::buffer_sequence_end(const_buffers); ++buffer) {
			consumed_total += _connection.parse(reinterpret_cast<const char*>(buffer->data()), buffer->size());
		}
```

The reason is that `_connection.parse` doesn't store or cache its input. So, if you would ever have real buffer sequence instead of single buffer one, then you needed to squash it before feeding into  `_connection.parse`.

matwey:
@corporategoth FYI, I've just force-pushed an updated version that is suitable for all Boost versions. It still relies on Boost implementation details but has a compile-time guard for the case when they change.

PR #536, opened 2024-11-07
Fix float double work

mpollov:
append Float and Double header work
fix out and in buffer work with float and double


mpollov:
@EmielBruijntjes whats up?

Issue #464, opened 2022-05-31
LibBoostAsioHandler does not detect connection timeouts when heartbeat is negotiated


ricardocrl:
Opposed to libev integration, libboostasio.h does not implement any mechanism to detect connection drops. 

At first, I tried using TcpHandler::onHeartbeat() to detect timeouts, but this is not sufficient because the broker does not send heartbeats if there is data flowing to us. Intercepting these events on top of boost::asio integration is not trivial or beauty.

When comparing to libev.h, this is solved with `onActive` being called and extending the timeout timer. So I decided to fix this by porting the solution into libboostasio.h, hooking the timer update in the most equivalent method `read_handler`. This is the PR for the changes proposed: #463.

On a separate note, I believe this code can still be refactored for a better integration with boost. I'm not fluent with boost or boost::asio, but I may suggest some changes with care.

Issue #479, opened 2022-08-31
Race condition leading to tampering with a socket we no longer own in libboostasio.h


devgs:
**Describe the bug**
Due to an async nature of `libboostasio.h` we **may get** into a situation when `Watcher` outlives the call to `monitor()` that uninstalls the descriptor watch (`flags == 0`). Leading to a use of a socket that was closed by caller of monitor (`TcpExtState::cleanup`) and, immediately after, created by another entity (in POSIX, descriptor numbers are reused).

Specifically, problematic here is a use of `_socket.release()` that unsubscribes the socket's descriptor from a reactor (epoll/kqueue/...) when the file descriptor is in use by another asio socket, basically making it inoperable.

**Expected behavior and actual behavior**
Don't tamper with file descriptors that you no longer own.

**Sample code**

It's really hard to provide a sample. It's always hard to reproduce a race. And, as usual, it only happens in production :). 
And obviously, to reproduce it you have to have at least two threads. In our case it's the other thread that manages to create a socket with a descriptor that was just `close()`-d by **amqpcpp**'s thread, but before it was `release()` in a `Wrapper`'s destructor.

It's easier to suggest a fix: either use a POSIX `dup()` or better: release the socket immediately, not relying for the `Wrapper`'s destructor to be invoked:

```
            // the watcher does already exist, but we no longer have to watch this watcher
            iter->second.release(); // <------- HERE
            _watchers.erase(iter);
```

Obviously, you also have to add a `release()` function to `Watcher`.

paolopas:
You described the problem well, but in my opinion the solution you propose doesn't completely solve it.

I say this because both the call to close the socket that occurs in `TcpExtState::cleanup` and the release that would occur in the `monitor` are executed by the same thread, which in the meantime could be suspended by the operating system scheduler.

Even ensuring that the `Watcher` is destroyed in the `monitor` (which is easy to do) would not solve the problem.

paolopas:
@devgs thanks for reporting a bug.
To understand what prevents the Watcher from being destroyed when `TcpExtState::cleanup()` runs, I traced the `use_count()` of the affected `shared_ptr`s.

I'm attaching the patch that instruments the current version (839f74f2e7cc0f3f03458ebf943d23fc1f8c5462) of `libboostasio.h` for debugging purposes. The test client is very simple, it just tries to connect to a _non-existent_ broker.

````c++
#include <boost/asio/io_context.hpp>
[...]

Issue #381, opened 2020-11-17
Support of ASIO standalone edition


nothisman:
Official website：http://think-async.com/Asio/AsioStandalone

Boost is so big to a native project and most of its frequently used methods are implemented by c++11/14, therefore, we're trying to remove boost from our project.

It will be nice if AMQP-CPP gives a support to standalone asio, which no longer a part of boost.

Thanks.

EmielBruijntjes:
Well, you can send in a pull request.

PR #550, opened 2025-12-02
libboostasio fully refactored

paolopas:
The reasons that pushed me to review the implementation of **libboostasio** are the following:

* uses deprecated `boost.asio` code, e.g. `deadline_timer`, `null_buffers`
* the header for `strand` is now `boost/asio/io_context_strand.hpp`
* virtual inheritance is being abused for `LibBoostAsioHandler` and `LibBoostAsioHandler::Watcher`, an unnecessary complication
* Heartbeat management is lacking and not in line with that offered by `libev`, in fact the provided implementation solves issue #464 by copying everything from `LibEvHandler`
* replace the use of `boost::bind` and `boost::function` with their standard library equivalents

Most of the work done is related to the callback code (i.e. `asio` completion handler):

* all callbacks now have the same prototype and are generated more consistently, generation is now centralized in the `make_handler` method
* callbacks now handle dependencies on this (`Watcher` instance) and the strand (`io_context::strand` instance) equally
* they are no longer invoked with `boost::system::errc::operation_canceled` if the strand has been destroyed as this serves no purpose
* They definitely shouldn't be called with `boost::asio::error::would_block`, see [boost::asio::posix::stream_descriptor documentation](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/reference/posix__basic_stream_descriptor/non_blocking/overload1.html)

**No changes have been made to the public interface of `LibBoostAsioHandler` except for the parameter (`uint16_t connection_timeout = 60`) added to the constructor (which has a default value anyway.)**

I just added a couple of `assert`s to handle dubious cases in the `monitor` and `onNegotiate` methods, I think the comments in the relevant lines are pretty explanatory. Perhaps the assert in `monitor` can be useful to prematurely intercept many of the problems related to an unsupported concurrency scheme or other weird issues.

Need more testing, of course...

### DIFF
--- a/README.md
+++ b/README.md
@@ -710,7 +710,7 @@ it has a couple of open issues.
 
 | TCP Handler Impl        | Header File Location   |  Sample File Location     |
 | ----------------------- | ---------------------- | ------------------------- |
-| Boost asio (io_service) | include/libboostasio.h | examples/libboostasio.cpp |  
+| Boost asio              | include/libboostasio.h | examples/libboostasio.cpp |
 | libev                   | include/libev.h        | examples/libev.cpp        |
 | libevent                | include/libevent.h     | examples/libevent.cpp     |
 | libuv                   | include/libuv.h        | examples/libuv.cpp        |
@@ -770,7 +770,7 @@ class MyTcpHandler : public AMQP::TcpHandler
 ````
 
 If you enable heartbeats, it is your own responsibility to ensure that the
-```connection->heartbeat()``` method is called at least once during this period,
+`connection->heartbeat()` method is called at least once during this period,
 or that you call one of the other channel or connection methods to send data
 over the connection. Heartbeats are sent by the server too, RabbitMQ also ensures
 that _some data_ is sent over the connection from the server to the client 
--- a/examples/libboostasio.cpp
+++ b/examples/libboostasio.cpp
@@ -5,15 +5,14 @@
  * 
  *  @author Gavin Smith <gavin.smith@coralbay.tv>
  *
- *  Compile with g++ -std=c++14 libboostasio.cpp -o boost_test -lpthread -lboost_system -lamqpcpp
+ *  Compile with
+ *  g++ -std=c++17 libboostasio.cpp -o boost_test -lpthread -ldl -lboost_system -lamqpcpp
  */
 
 /**
  *  Dependencies
  */
-#include <boost/asio/io_service.hpp>
-#include <boost/asio/strand.hpp>
-#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/io_context.hpp>
 
 
 #include <amqpcpp.h>
@@ -27,30 +26,41 @@ int main()
 {
 
     // access to the boost asio handler
-    // note: we suggest use of 2 threads - normally one is fin (we are simply demonstrating thread safety).
-    boost::asio::io_service service(4);
+    boost::asio::io_context context;
 
-    // handler for libev
-    AMQP::LibBoostAsioHandler handler(service);
-    
+    // handler for libboostasio
+    AMQP::LibBoostAsioHandler handler(context);
+
     // make a connection
     AMQP::TcpConnection connection(&handler, AMQP::Address("amqp://guest:guest@localhost/"));
-    
+
     // we need a channel too
     AMQP::TcpChannel channel(&connection);
-    
+
+    channel.onReady([]() {
+        std::cout << "channel ready" << std::endl;
+            });
+    channel.onError([](const char *message) {
+        std::cout << "channel error: " << message << std::endl;
+            });
+
     // create a temporary queue
     channel.declareQueue(AMQP::exclusive).onSuccess([&connection](const std::string &name, uint32_t messagecount, uint32_t consumercount) {
-        
+
         // report the name of the temporary queue
         std::cout << "declared queue " << name << std::endl;
-        
+
         // now we can close the connection
         connection.close();
+
+    }).onError([](const char *message) {
+
+        // something went wrong creating the queue (or even before)
+        std::cerr << "declareQueue error: " << message << std::endl;
     });
-    
+
     // run the handler
-    // a t the moment, one will need SIGINT to stop.  In time, should add signal handling through boost API.
-    return service.run();
+    // at the moment, one will need SIGINT to stop.  In time, should add signal handling through boost API.
+    return context.run();
 }
 
--- a/include/amqpcpp/libboostasio.h
+++ b/include/amqpcpp/libboostasio.h
@@ -379,7 +379,7 @@ protected:
             _read = ((events & AMQP::readable) != 0);
 
             // Read requested but no read pending?
-            if (_read && (_read != old_read) && !_read_pending)
+            if (_read && ((_read != old_read) || !_read_pending))
             {
                 _read_pending = true;
 
@@ -393,7 +393,7 @@ protected:
             _write = ((events & AMQP::writable) != 0);
 
             // Write requested but no write pending?
-            if (_write && (_write != old_write) && !_write_pending)
+            if (_write && ((_write != old_write) || !_write_pending))
             {
                 _write_pending = true;
 
--- a/include/amqpcpp/libboostasio.h
+++ b/include/amqpcpp/libboostasio.h
@@ -1,17 +1,18 @@
 /**
  *  LibBoostAsio.h
  *
- *  Implementation for the AMQP::TcpHandler for boost::asio. You can use this class 
- *  instead of a AMQP::TcpHandler class, just pass the boost asio service to the 
- *  constructor and you're all set.  See tests/libboostasio.cpp for example.
+ *  Implementation for the AMQP::TcpHandler for boost::asio. You can use this class
+ *  instead of a AMQP::TcpHandler class, just pass the boost asio io_context to the
+ *  constructor and you're all set.  See examples/libboostasio.cpp for an example.
  *
- *  Watch out: this class was not implemented or reviewed by the original author of 
+ *  Watch out: this class was not implemented or reviewed by the original author of
  *  AMQP-CPP. However, we do get a lot of questions and issues from users of this class,
  *  so we cannot guarantee its quality. If you run into such issues too, it might be
  *  better to implement your own handler that interact with boost.
  *
  *
  *  @author Gavin Smith <gavin.smith@coralbay.tv>
+ *  @author Paolo Pastori
  */
 
 
@@ -24,23 +25,17 @@
  *  Dependencies
  */
 #include <memory>
+#include <chrono>
+#include <functional>
+#include <cassert>
 
 #include <boost/asio/io_context.hpp>
-#include <boost/asio/strand.hpp>
-#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/io_context_strand.hpp>
+#include <boost/asio/steady_timer.hpp>
 #include <boost/asio/posix/stream_descriptor.hpp>
 #include <boost/asio/dispatch.hpp>
-#include <boost/bind/bind.hpp>
-#include <boost/function.hpp>
 
 #include "amqpcpp/linux_tcp.h"
-
-// C++17 has 'weak_from_this()' support.
-#if __cplusplus >= 201701L
-#define PTR_FROM_THIS(T) weak_from_this()
-#else
-#define PTR_FROM_THIS(T) std::weak_ptr<T>(shared_from_this())
-#endif
 
 /**
  *  Set up namespace
@@ -51,14 +46,14 @@ namespace AMQP {
  *  Class definition
  *  @note Because of a limitation on Windows, this will only work on POSIX based systems - see https://github.com/chriskohlhoff/asio/issues/70
  */
-class LibBoostAsioHandler : public virtual TcpHandler
+class LibBoostAsioHandler : public TcpHandler
 {
 protected:
 
     /**
      *  Helper class that wraps a boost io_context socket monitor.
      */
-    class Watcher : public virtual std::enable_shared_from_this<Watcher>
+    class Watcher : public std::enable_shared_from_this<Watcher>
     {
     private:
 
@@ -72,7 +67,7 @@ protected:
 
         /**
          *  The boost asio io_context::strand managed pointer.
-         *  @var class std::shared_ptr<boost::asio::io_context>
+         *  @var std::weak_ptr<boost::asio::io_context::strand>
          */
         strand_weak_ptr _wpstrand;
 
@@ -84,10 +79,41 @@ protected:
         boost::asio::posix::stream_descriptor _socket;
 
         /**
-         *  The boost asynchronous deadline timer.
-         *  @var class boost::asio::deadline_timer
+         *  The boost asynchronous timer.
+         *
+         *  Used for monitoring of both server connection timeout condition
+         *  (at initial connection stage) and heartbeat (client and server).
+         *  @var class boost::asio::steady_timer
          */
-        boost::asio::deadline_timer _timer;
+        boost::asio::steady_timer _timer;
+
+        /**
+         *  Timeout after which the connection is no longer considered alive.
+         *  A heartbeat must be sent every _timeout / 2 seconds.
+         *  Value zero means heartbeats are disabled, or not yet negotiated.
+         *  @var uint16_t
+         */
+        uint16_t _timeout = 0;
+
+        /**
+         *  AMQP Server connection timeout setting (seconds).
+         *  @var uint16_t
+         */
+        uint16_t _connection_timeout = 0;
+
+        using steady_time_point = std::chrono::time_point<std::chrono::steady_clock>;
+
+        /**
+         *  When should we send the next heartbeat?
+         *  @var std::chrono::time_point<std::chrono::steady_clock>
+         */
+        steady_time_point _next;
+
+        /**
+         *  When does the connection expire / was the server idle for too long?
+         *  @var std::chrono::time_point<std::chrono::steady_clock>
+         */
+        steady_time_point _expire;
 
         /**
          *  A boolean that indicates if the watcher is monitoring for read events.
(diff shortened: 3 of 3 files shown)


paolopas:
As far as heartbeat management is concerned, in my opinion there is still one thing missing.
On #510 @EmielBruijntjes speaking of `LibEv` said:
> * We also take into account that no heartbeats have to be sent if we already sent some other kind of data

I haven't been able to find where this feature is actually implemented. If it isn't implemented in the library, I can tell you that `LibEvHandler::Wrapper::_next` isn't updated when sending data to the broker.
What is actually done, in the `Wrapper::onActive` method is updating `LibEvHandler::Wrapper::_expire`

https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/839f74f2e7cc0f3f03458ebf943d23fc1f8c5462/include/amqpcpp/libev.h#L286-L299

Maybe a line of code like

````c++
if (_timeout != 0 && (events & EV_WRITE)) _next = ev_now(_loop) + std::max(_timeout / 2, 1);
````
or a better code layout that avoids duplication (`_timeout` checking and `ev_now(_loop)` calls) could have the desired effect.

The same feature can obviously also be implemented in the `libboostasio`.

**Comment by paolopas** [comment]
I should also point out that issue #479 is easily fixable, although the reason I'm mentioning it here may not seem obvious.
@devgs is absolutely right and described the scenario well, and the proposed solution should work too.

With my proposed implementation (and even the current one) there is no apparent reason why a `Watcher` should outlive `TcpExtState::cleanup()`, there aren't `shared_ptr`s around (in AMQP_CPP code) keeping it alive. In the absence of evidence, one might assume that the delayed destruction of the `Watcher` is due to pending `boost` callbacks (read/write/timeout, which could have a significant impact and needs to be investigated.)

Well, releasing the socket certainly prevents the race condition described in the issue, and it also eliminates any callbacks related to pending operations on the socket, see [posix::basic_stream_descriptor::release documentation](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/reference/posix__basic_stream_descriptor/release.html)
**but not that related to the timer**!

So when you want to destroy the Watcher, you might need to release the socket and also cancel the timer, which is essentially equivalent to calling the destructor. Since both of these operations can be repeated without causing problems (subsequent releases do nothing), you can write a method that can also be called again by the destructor.

All this to finally say that clearing the `_read `and `_write` flags that occurs in the destructor (in addition to the other operations mentioned) is **absolutely useless** (in fact, it's the only thing I regretted leaving as it was), since when the
destructor is called, _there certainly can't be any `boost` callback that could use that data_.

**Comment by paolopas** [comment]
As expected, the problem of the delayed destruction of the `Watcher` is not due to `boost`, nor to the pending callbacks (waiting to be executed), but is in fact a bug in `libboostasio`.

I discuss the details in issue #479, but here I'll simply say that in my tests, the implementation proposed with this PR behaves exactly like the current version (except for being faster), so the proposed version also needs the same fix.

#479 demonstrates how important it is for callbacks to be as simple as possible. My version (which eliminates the need to include a `lock()` on a `weak_ptr`) is certainly to be appreciated, in addition to the greater simplicity, which I believe is
more important for maintenance.

**Comment by paolopas** [comment]
@EmielBruijntjes,
I hope you find the time one day to take a look at all this stuff. I understand no one asked me to do so, but the quality of this component is indeed quite far from the rest of the library. Of course, it can be done even better...
And if I can be of any help, I usually do so voluntarily. But feel free to ask.

**Comment by paolopas** [comment]
Ok, sorry for the mess with the comments but I wanted the code to be clear. I think that's all for now, it's time to do more testing.

Even though `libboostasio` is now faster (and above all more robust and clear) there is still a lot of room for improvement, here are some ideas (maybe @zerodefect might be interested):

## expose `strand`

Users of this handler may want to run some code in the same `io_contexct::strand` used by the handler. This requires adding a method to the public interface of `LibBoostAsioHandler`.

## improve performance

There is really a lot of room for improvement here.

### Abandon the use of managed pointers to ensure that the strand or watcher pointer does not dangle

The lifecycle of a `Watcher` should be guaranteed only by canceling all callbacks in the destructor (as it already is). Everything else can be done with the usual observer pattern, that is reusing `AMQP::Watchable` and `AMQP::Monitor`.

### abandon the use of `Watcher::make_handler`

Yes, while this method I wrote to clean up the current implementation improves performance quite a bit, I'm convinced that this style of asynchronous programming is inherently inefficient. There should be no need to generate a wrapper around the method that needs to complete an asynchronous operation (e.g. _every time the socket becomes readable/writable_) when that method does not change.

Why not use a coroutine instead? Here's a sketch of this idea:

```c++
#include <boost/asio/coroutine.hpp>
#include <boost/asio/yield.hpp>

class Watcher:
private:

    boost::asio::posix::stream_descriptor _socket;

    ...

    boost::asio::coroutine _rd_coro;

    bool _read{false};
    bool _read_pending{false};

public:

    // initiating call (to be called by LibBoostAsioHandler::events()/LibBoostAsioHandler::monitor())
    void async_reading(bool want_read)
    {
        bool old_read = _read;
        _read = want_read;
        if (_read && ((_read != old_read) || !_read_pending)) // see #424 
        {
            _read_pending = true;
            _rd_coro = boost::asio::coroutine(); // coroutine reset
            _socket.async_wait(
                boost::asio::posix::stream_descriptor::wait_read,
                [this](const boost::system::error_code& ec) { async_reading(ec); }); // strand omitted
        }
    }

private:

    // working call (the coroutine)
    void async_reading(const boost::system::error_code& ec)
    {
        _read_pending = false;
        reenter (_rd_coro) while (!ec && _read)
        {
            // do heartbeat stuff ...

            // do connection->process ...

            if (_socket.is_open())
            {
                _read_pending = true;
                yield _socket.async_wait(
                    boost::asio::posix::stream_descriptor::wait_read,
                    [this](const boost::system::error_code& ec) { async_reading(ec); }); // strand omitted
            }
        }
    }

   ...

#include <boost/asio/unyield.hpp>
```

paolopas:
I know, this PR is getting too long...
But I saw #424 late, and it seems too easy to fix not to do it.
So I modified `Watcher::events` and the above example (the coroutine one) to fix the problem.

paolopas:
I'm still not sure that the condition that should solve #424 is 100% reliable.
Well, I wouldn't call it a thorough fix; the solution proposed by @svyatogor addresses the problem in a more structured way. However, as a workaround, it should work well.

zerodefect:
Thanks for this contribution. This PR covers a significant amount of ground.

Because this project is used in production, we need to be extremely cautious about introducing regressions. Reviewing a change of this magnitude in one go makes it difficult to spot potential edge cases.

Could you please split this into smaller, atomic PRs? I recommend an iterative approach: start with the code cleanup/refactoring first, and then add the functional changes in subsequent PRs. I would be happy to prioritize reviewing those smaller PRs to get them merged quickly.

paolopas:
I think now #424 is really fixed.

@zerodefect I understand well, but in the meantime here's the whole story. I myself want to do several more tests.

Meanwhile, for those who want to have a comprehensive idea, here's where we are

> [!IMPORTANT]
> [**libboostasio_2025-12-07.zip**](https://github.com/user-attachments/files/24183007/libboostasio_2025-12-07.zip)

paolopas:
But why did you ask me to do this if you wanted to do it yourself?
> 
> Could you please split this into smaller, atomic PRs? I recommend an iterative approach: start with the code cleanup/refactoring first, and then add the functional changes in subsequent PRs. I would be happy to prioritize reviewing those smaller PRs to get them merged quickly.

zerodefect:
Your response came across as dismissive. You responded with:

>  I understand well, but in the meantime here's the whole story. I myself want to do several more tests.

You understand, but you don't indicate that you are going to do what I suggested. It's not my repo. I don't make the decisions around here.

paolopas:
then it means I have to thank @EmielBruijntjes for this... I'm still happy that in the end you understood the value of what was already given away.

paolopas:
Yes, @zerodefect, okay, we didn't understand each other. But it's the reason for _all this haste_ that bothers me. Considering that @devg (#479) waited more than 3 years without a response (despite having understood and explained the problem better, as well as having served up the solution on a silver platter). And let's not talk about @svyatogor (#424) who has been waiting for more than 4 years for someone to pay attention to him.
I didn't write libboostasio, but that doesn't stop me from looking inside, understanding where the problems are, and maybe even finding solutions (and maybe in less than 5 days). It's not my repository of course...

But these are the facts:

+ #464 fixed, #479 fixed, #424 surely too (not yet tested well), #381 also involved (just discovered), plus that little heartbeat problem with `libev` (and this is not `libboostasio`), and many good ideas for the future (note I said future).
+ The code performs better and above all it is much simpler and clearer.
+ All the tests I'm running tell me that everything is fine and there are no regressions and the only thing I would add to the code now are a couple of comments

Now the simple question is @EmielBruijntjes **should I continue to produce PR like #553 or have you decided to do it yourself?**

zerodefect:
Let me be clear - I understood you. I  didn't like your approach. If there's an issue spotted, does Emiel revert all your changes? We're back to square one.  Effectively, nothing has then been updated.

I have an interest in this project. If it goes wrong, it costs me.

As I said, there is some good stuff in there like the heartbeat(ing). Can your changes be used? Sure. Let's layer changes. I have other changes waiting in the wings. It's always been difficult to get changes merged into this repo.

paolopas:
I don't understand what the problem is with the revert. I didn't like your approach either. 1-1.
If you really have an interest in this project then you shouldn't let that code circulate, it's not a good calling card. The first time I saw the carousel of calls to return a handler I trembled...
@zerodefect, don't worry, I do all the tests I can, and in fact if you remember I told you I wanted to do more, I'm here to help, #553 should prove it to you. But you can see a lot from the code, and taking a look at the diffs is worth more than a lot of chatter, and if you don't understand much from a diff, **git** allows you to see the whole picture very easily. After all, that's why we're here, right? As I already told Emiel, _feel free to ask_, and even if he doesn't want to have anything to do with boost, it will become more and more widespread and used, that's a fact.

zerodefect:
Hi Paolo, I'm not here to score points nor to seek public appreciation or approval. I was just sharing my concerns. Go well.

paolopas:
With the #558, I've finished factorizing this work, at least up to this point. The code is basically the same (except for one condition I improved), but it looks significantly better. I'm also adding the current version here for anyone who wants to try it.

> [!IMPORTANT]
> [**libboostasio_2025-12-13.zip**](https://github.com/user-attachments/files/24183085/libboostasio_2025-12-13.zip)

In the meantime, I've done all the testing I could, and this has led me to the following conclusions, some of which I've partially expressed in a note I left at the bottom of the header, but I have to update that note to talk about the elephant in the china shop. But let's save that for last.

1. A slightly more sophisticated test than the current one should be provided, which could also be used for testing purposes. I've written several, but something similar to libev's could already be considered a major improvement. I reserve the right to dedicate a PR to this purpose in the future.

2. The handler lacks a mechanism that allows the user to force the removal of `Watcher`s. There will definitely be a need for a new method in the handler.

I have often encountered processes that would not terminate because zombie watchers remained in the reactor queue but were attached to an unresponsive filedescriptor. And judging by the issues from several users, it is a frequent case, which pushes many (wrongly) towards a multi-threaded solution (the elephant.)

This can happen under different conditions and ensuring that the reactor (boost `io_context` so to speak) is flushed along each path is non-trivial.

For example, in one of the tests I use a publisher that does not implement the `onBlocked` and while it is running I set the rabbit's `rabbitmqctl set_vm_memory_high_watermark absolute` to 0. The server blocks it but the client does not understand what is happening and when I try to manually stop it I have to do a kill. The problem is that even if I had implemented `onBlocked`, I would not have been able to empty the reactor's queue without specific handler support, or at least not in a clean way. At the moment, it's hard to know what other use cases this might actually be needed for, so I propose adding it as a protected method. Perhaps I can also modify the example to demonstrate its use in the most common case.

I have some doubts about what the interface of the single method to be added could be, considering that the user of the `TcpHandler` thinks in terms of `TcpConnection` but the handler thinks in terms of filedescriptor, it might sound like this
```c++
protected:
    /**
     *  Stop handling connection(s).
     *
     *  Note that you cannot continue using the connection(s) after calling this method.
     *  @param  pconn  The TcpConnection* to release (all by default).
     *  @return bool   If there was a release.
     */
    bool release(const TcpConnection *pconn = nullptr)
    {
        if (pconn == nullptr)
        {
            // close all watchers
            for (auto &wp : _watchers) wp.second->close();
            auto b = _watchers.begin();
            return b != _watchers.erase(b, _watchers.end());
        }
        else
        {
            int fd = pconn->fileno();
            // is the connection already closed?
            if (fd != -1) {
                // close matching watcher
                auto iter = _watchers.find(fd);
                if (iter != _watchers.end())
                {
                    iter->second->close();
                    _watchers.erase(iter);
                    return true;
                }
            }
            return false;
        }
    }
```
Which obviously should also be called during the destruction phase, always with the aim of preventing unwanted deadlocks.
I did some tests and it works, except that I have to remove the assert that I had put in the `monitor` method, in case of forced release of a connection it is only the handler that forgets about it but the library does not and sooner or later tries to unregister it. Actually that assert was the only thing that worried me a little.
I look forward to your feedback on this @EmielBruijntjes.

We finally got to the elephant.

## about libboostasio and thread safety

AMQP-CPP is not thread safe, we cannot repeat this too many times.

But I think we should definitely give credit to Gavin, the original author of this handler, for the intuition that with the help of the boost reactor it could get pretty close.

The reactor ensures that queued callbacks are executed exclusively by threads that are running the context (`io_context::run()`) and the `io_context::strand` ensures that each callback is executed sequentially and every other thread in the pool sees its effects.

So `boost` magically makes `AMQP-CPP` thread safe?

**_if all the threads involved are running the context then all the calls to AMQP-CPP originate from callbacks (at least as far as this handler is concerned) and therefore there are no problems
BUT
if a thread interacts with the library without going through the reactor (for example it directly invoke `publish`) then expect an indeterminate behavior_**.

This is why among the proposed improvements for the future I said that the handler should expose the `strand`.

paolopas:
# Why we need to eliminate managed ptrs
In #562 (ddc1662a25bb6f21813fc293ab9d4eba1f75e3d7) I removed all the managed ptrs that pointed to the strand from `AMQP::LibBoostAsioHandler`. The first was created during construction of the handler, which in the `_strand` member held a shared_ptr for the dynamically allocated object.
The first time I saw it I wondered why it was written that way, and I still haven't found an answer. It certainly hinders our plans to expose the strand (which was created for the handler's exclusive use) to users who need it. Certainly, passing a shared_ptr in user-space (pardon the somewhat specious expression) is certainly not the best possible interface; perhaps a simple reference would be better.
But let's move on. The manager doesn't use the strand directly, its `Watcher` children need it to send callbacks.
Note that all callbacks are executed with `dispatch` and not deferred with `post`. This is to ensure that the queueing order in the reactor is preserved. And practically all the code in `libboostasio.h` (except the handler constructor) is executed through a callback in the event loop. Since the strand ensures that callbacks cannot overlap, even if they may execute in different threads, the resulting effect is to emulate a single-threaded process.
So, for example, there's no danger that while this virtual thread is executing the `make_callback` (which is where all the handler drawing errors have accumulated up to now), it could be interrupted by the same virtual thread, which could start destroying not only the watcher (which can happen during normal operation) but even the parent!
Now I understand that this reasoning is difficult to follow, and I'm not good at explaining myself. But let's try to look at it differently.
Let's think about the order of construction/destruction of objects. Well, when the handler is constructed, the strand attribute comes first, followed by the map with the children (watchers). Yes, I know, they can keep themselves alive because they maintain a shared_ptr at `this`, but what matters is that the children destruction will certainly be attempted first, and THEN the strand.
Now, since we can write the handler however we want, no one is stopping us from forcing the destruction of a children whenever we want. And it's simple, because the destructor simply fires all the events enough to make the watcher exit any callbacks that could allow it to survive. The fix for #479 (see 5d03768ea1bc92b071c950e00844475fc68f28ac) for example did just that.
So rather than filling the code with managed ptrs (with significant performance costs and some risk of unwanted effects) out of fear that a pointer might dangle, we should understand that we're in a scenario where this eventuality can easily be avoided in a less expensive way. I have always tried to make the various handlers that execute callbacks as simple (and responsive) as possible, ensuring that they check as soon as possible whether it's time to abandon execution, perhaps because the socket was closed while they were waiting to be scheduled (or running `connection->process()`) ...
In short, what ensures that in the `make_handler` method we do not end up with a dying parent is precisely the fact that (by design) we children must die first! So I happily threw away all the shared ptrs that the children wanted to keep on the parent's strand and replaced them with a pointer to the parent, a simple good old pointer.
And if it's not clear yet, I'll say it another way.
Imagine we're executing the watcher's destructor, and there are still some callbacks (from the same watcher) in the reactor's queue; now, if the destructor requests their deletion, since the reactor calls them immediately with an error (abort), if they are written carefully, they immediately return control to the destructor, which can continue. Problems arise when the watcher's destructor isn't called, then the watcher can survive longer than it should, and this can only happen thanks to the damned shared ptrs we've put on guard for `this` to sleep soundly. 
So instead of adding more, it would be much safer to remove all of them. And if you want to be sure there are no queued callbacks (which, even if they don't keep the watcher alive, can still have other unwanted effects), just be sure to delete them.
_This is why I resolve the issue by saying that the guarantee that the parent cannot die before the children must be written into the children's destructor and in the care that must be taken to complete the work left by the children when it is certain that it is no longer needed_.
For those wondering what happens to the callbacks that are running when the watcher is destroyed, here is a simple program that demonstrates that there is nothing to fear [**proof23.zip**](https://github.com/user-attachments/files/24358377/proof23.zip).

I just started throwing away shared ptr...

paolopas:
I'm done with #566, I promise. Also because I fixed everything that was wrong except for one thing that, however, there's nothing that can be done.
In the meantime, I also did some experiments to explore the feasibility of the coroutine-based implementation and I realized that it is not practicable (and not even convenient as long as we are forced by boost to have to wrap every callback.) The problem with coroutines is that while a callback can be easily canceled, forcing the coroutine to exit is virtually impossible.

The handler and its children now finally have the minimal architecture that allows them to function efficiently. Globally we have gone from this
```mermaid
classDiagram
LibBoostAsioHandler : SharedPtr~Strand~ _strand
LibBoostAsioHandler : Map~int SharedPtr~Watcher~~ _watchers
LibBoostAsioHandler *-- Watcher : _watchers
class enable_shared_from_this~Watcher~
Watcher --|> enable_shared_from_this~Watcher~
Watcher : WeakPtr~Strand~ _wpstrand
Watcher o-- LibBoostAsioHandler : _wpstrand
```
to this
```mermaid
classDiagram
LibBoostAsioHandler : Strand _strand
LibBoostAsioHandler : Map~int UniquePtr~Watcher~~ _watchers
LibBoostAsioHandler *-- Watcher : _watchers
Watcher : Strand & _strand
Watcher --> LibBoostAsioHandler : _strand
```
It may seem like a small change but it has important effects on how it works.
I thank everyone for their patience and I attach the latest version

> [!IMPORTANT]
> [**libboostasio_2025-12-23.zip**](https://github.com/user-attachments/files/24314828/libboostasio_2025-12-23.zip)

PR #551, opened 2025-12-07
Refactor: Modernize LibBoostAsioHandler to C++17 syntax (pt 1)

zerodefect:
This commit updates the Boost Asio handler implementation to remove dependency on legacy Boost libraries in favor of modern C++17 standard features.

Changes include:
- Replaced all instances of `boost::bind` with native C++ lambdas.
- Removed `boost/bind/bind.hpp` dependency.
- Replaced `PTR_FROM_THIS` macro with standard `weak_from_this()`.
- Standardized callback types to use `std::function` instead of `boost::function`.
- Use `std::move()` semantics to reduce memory allocation(s).

PR #552, opened 2025-12-08
Refactor: Replace LibBoostAsioHandler deadline_timer with steady_timer (pt 2)

zerodefect:
This commit replaces `boost::asio::deadline_timer` with `boost::asio::steady_timer`.

The previous implementation relied on the system wall clock (`boost::posix_time`). This caused issues where heartbeat timers could hang or fire incorrectly if the system time was adjusted (e.g., via NTP or Daylight Savings).

The new implementation uses `std::chrono::steady_clock` (monotonic), ensuring timers fire at consistent intervals regardless of system time changes.

### DIFF
--- a/include/amqpcpp/libboostasio.h
+++ b/include/amqpcpp/libboostasio.h
@@ -23,11 +23,12 @@
 /**
  *  Dependencies
  */
+#include <chrono>
 #include <memory>
 
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/strand.hpp>
-#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/steady_timer.hpp>
 #include <boost/asio/posix/stream_descriptor.hpp>
 #include <boost/asio/dispatch.hpp>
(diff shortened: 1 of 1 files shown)


zerodefect:
PR #551 should be merged before this one. PR 551 may necessitate conflict resolution thereafter.

paolopas:
I probably didn't understand how you intend to proceed, if we both try to factorize #550 we will probably waste a lot of time.
See #553  and your previous comment  https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/550#issuecomment-3622830360

zerodefect:
I feel your changes are too ambitious - too many changes at once. 

If one needs to do a git bisect, it becomes difficult to nail down a problem if too many things are globbed.

zerodefect:
For those interested, I've created a new repository which encapsulates an improved/updated implementation of LibBoostAsioHandler (aka LibBoostAsioHandler2).

https://github.com/coralbay-tv/amqp-cpp-boost2

PRs welcome :)

PR #554, opened 2025-12-09
Refactor: Encapsulate file descriptor in LibBoostAsioHandler Watcher class (pt 3)

zerodefect:
This commit moves the file descriptor (`fd`) into a member variable `_fd`  within the `Watcher` class.

Previously, `fd` was passed as an argument through every internal handler and captured in every lambda. Since the `Watcher` owns the socket, passing the `fd` explicitly was redundant. This change simplifies the function signatures and reduces the lambda capture overhead.

### DIFF
--- a/include/amqpcpp/libboostasio.h
+++ b/include/amqpcpp/libboostasio.h
@@ -105,6 +105,12 @@ protected:
          *  @var _read True if read is pending else false.
          */
         bool _write_pending{false};
+        
+        /**
+         *  The socket descriptor that is being monitored.
+         *  @var _fd The OS's underlying socket descriptor.
+         */
+        int _fd;
 
         using handler_cb = std::function<void(boost::system::error_code, std::size_t)>;
         using io_handler = std::function<void(const boost::system::error_code&, const std::size_t)>;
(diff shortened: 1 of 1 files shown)


paolopas:
Hi Zero,
Looking at the diff, there are a few things I think you should consider more carefully:
1. for example in `get_dispatch_wrapper` you changed the error from `operation_canceled` to `operation_aborted`, right that's exactly what `boost` does on callbacks when they are cancelled. This seems like a rather unlikely scenario to me, if the handler cancels all callbacks during destruction.
2. It's not accurate to say that the watcher has the socket. What the watcher has is a `boost::asio::posix::stream_descriptor`, which is initialized on the same file descriptor as the REAL socket that someone else has, just so it can observe it. They are two different objects; both can be closed or released independently.
3. What never changes, however, is the connection pointer; perhaps that could have been saved in the watcher.

But what I think is that right now it's convenient to have both immediately available as parameters, and passing a pointer and an integer isn't a huge waste of resources. But above all, I think that before optimizing code, you need to clean up all the problems it has (and believe me, there are quite a few here), otherwise you'll just unnecessarily complicate the overall picture.

zerodefect:
For those interested, I've created a new repository which encapsulates an improved/updated implementation of LibBoostAsioHandler (aka LibBoostAsioHandler2). 

https://github.com/coralbay-tv/amqp-cpp-boost2

PRs welcome :)

paolopas:
@zerodefect, I think I explained it better here https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/550#issuecomment-3678053222 how the handler works and how one should proceed with particular attention to the method that in your code is still called `get_dispatch_wrapper` but which instead I call `make_handler` into my version. I'm sure it will help you write a better version.

zerodefect:
@paolopas Thanks for the review and the feedback. I appreciate you taking the time to look through the diffs.

Regarding the error codes and ownership semantics, you raise some interesting architectural points. As I mentioned, I've moved the active development of this modern implementation to a dedicated repository (amqp-cpp-boost2) to allow for more significant refactoring without disrupting the legacy codebase here.

I'd be happy to continue the discussion there—if you have time, please feel free to open an issue or PR on that repo so we can track these specific improvements separately. For this PR, I'm aiming to keep the changes focused on the iterative cleanup we've already established.

paolopas:
I'm sorry, but I see that you're badly copying the code you accessed, reintroducing errors that I had expertly eliminated, so it doesn't matter where you do it, in my opinion it's more important how. No one's "disrupting legacy code" here, no interface has changed, I'm just fixing all its many flaws. Problems that have waited years before I saw them...

zerodefect:
I understand you have a strong preference for your own implementation, but I am satisfied with the stability and trade-offs of this approach. I am freezing the scope of this PR here. If you believe your implementation is superior, I encourage you to submit your own PR to the maintainers for review. I will not be debating this further.

PR #562, opened 2025-12-18
Refactoring libboostasio (part 5)

paolopas:
> [!NOTE]
> This is the fifth child, the code is branched from #558 (fourth child), so this one accumulate the changes.

Each commit has been checked for regressions.

+ 38c650c7bdccbc29abb12d731173d077e5d0ceeb disclaimer notice added, final notes updated too
+ b461f496e5b90fbc761d43e160df71da6f04a0a0 new `release` method (protected) and removal of the `assert` in the `monitor` method, as it is now possible for managed connections to be released early (as described in https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/550#issuecomment-3650308626)
+ 6e152ec250f2c90221603c6f2e4c8b716e64e899 minor optimizations in watcher creation and renamed `Watcher::events` method to `set_event_mask`
+ ddc1662a25bb6f21813fc293ab9d4eba1f75e3d7 watcher now holds a plain pointer to the parent and no longer need the managed pointers for the strand. This makes creating callbacks a little more efficient, as you no longer have to resort to the usual juggling of the weak/shared ptrs. pair to ensure you don't have dangling pointers. I explained here why this modification, in addition to having better performance, _is absolutely safe_  https://github.com/CopernicaMarketingSoftware/AMQP-CPP/pull/550#issuecomment-3678053222
+ 6143b87c87dac38ace0a28b2dd7f362cc2712c82 the example of using liboostasio has been completely rewritten, and as always it is inspired by the one relating to libev. In addition to being equipped with a disclaimer, it documents the use of the handler in more detail.
+ 4b4d5924fc45c075d97a7c479296d4a56a0f6bcb simplified libboostasio test

I'm attaching the current version of the handler (with example test included) for anyone curious to read or experiment.

> [!IMPORTANT]
> [**libboostasio_2025-12-19.zip**](https://github.com/user-attachments/files/24250093/libboostasio_2025-12-19.zip)

Issue #547, opened 2025-10-10
won't publish until heartbeat received for SSL connection


pauliuscoinpoker:
**Describe the bug**
library won't publish a message unless heartbeat is received for SSL connection

**Expected behavior and actual behavior**
should publish instantly

**Sample code**
```c++
#include <amqpcpp.h>
#include <amqpcpp/libboostasio.h>
#include <boost/asio.hpp>
#include <openssl/ssl.h>

// .. some helpers omitted

int main() {
    OPENSSL_init_ssl(0, nullptr);

    boost::asio::io_context io;

    Handler handler(io);

    AMQP::TcpConnection connection(&handler, AMQP::Address("amqps://" HOST_NAME "/" VHOST_NAME));

    AMQP::TcpChannel channel(&connection);
    channel.declareQueue(QUEUE_NAME);

    auto relChannel = AMQP::Reliable(channel);
    auto consChannel = AMQP::TcpChannel(&connection);

    consChannel.consume(QUEUE_NAME, AMQP::noack).onReceived([&consChannel](const AMQP::Message& msg, uint64_t tag, bool) {
        Log() << "--- consumed: " << msg.body();
    });

    ReocurringTimer t(io, [&relChannel]() {
        static int i{};
        auto msg = std::format("m{}", ++i);
        Log() << "--- publishing: " << msg;
        relChannel.publish("", QUEUE_NAME, msg).onAck([msg]() { Log() << "publish acked: " << msg; });
    }, std::chrono::seconds(2));

    io.run();
}
```

output:

```
[11:06:37] --- publishing: m1
[11:06:39] --- publishing: m2
[11:06:40] ** heartbeat received
[11:06:40] publish acked: m1
[11:06:40] publish acked: m2
[11:06:40] --- consumed: m1
[11:06:40] --- consumed: m2
[11:06:41] --- publishing: m3
[11:06:43] --- publishing: m4
[11:06:45] --- publishing: m5
[11:06:47] --- publishing: m6
[11:06:49] --- publishing: m7
[11:06:50] ** heartbeat received
[11:06:50] --- consumed: m3
[11:06:50] publish acked: m3
[11:06:50] publish acked: m4
[11:06:50] publish acked: m5
[11:06:50] --- consumed: m4
[11:06:50] --- consumed: m5
[11:06:50] --- consumed: m6
[11:06:50] --- consumed: m7
[11:06:50] publish acked: m6
[11:06:50] publish acked: m7
[...]

PR #566, opened 2025-12-23
Refactoring libboostasio (part 6)

paolopas:
> [!NOTE]
> This is the sixth and **last** child, the code is branched from #562 (fifth child), so this one accumulate the changes.

Each commit has been checked for regressions.

+ 1870a70a8c9b84af1276a1aef72e96efbec6d375 finally, complete liberation from the useless shared_ptr is achieved, 23 commits ago, I never thought I'd see this... Now not only does the `Watcher` no longer have to inherit from `enable_shared_from_this`, but more importantly, the `make_handler` avoids creating another pair of weak/shared ptr to guard `this` (which previously was done at every callback invocation, i.e. every time the socket becomes readable or writable, even just to receive 1 byte!) But `make_handler` is now smarter for another reason too: in case of errors from boost it exits immediately without calling the callback at all. As I already explained, this, together with the cancellation of the callbacks by the destructor, allows you not to have to worry too much about the dangling of the `this`. Finally, the handler now stores children using `unique_ptrs` instead of `shared_ptrs`. Simple pointers could have been used, but I preferred to stay aligned with the other handlers. Now, when you erase the child in the map, the object is certainly destroyed. There are no delays of any kind, and no need to be careful to avoid releasing sockets too late. In short, everything works as expected, with no surprises. This is the greatest benefit for anyone who reads or maintains the code. This and a certain amount of appropriate comments added where needed.
+ 1c11cd9f4d9b2a23f7b22f2785da593c7a659ed3 given that callback methods are never called when boost reports an error, it is clear that they do not need to have the error as a parameter.
+ b1355f1c5b47136385884ef8a75a238c203363d1 I replaced the pointer to the parent in the watchers with a reference to the `strand`, after all they don't ask much else from their parent.

As usual, for the more impatient among you, I'm leaving this latest version below. If the one that existed before I started working on it _was in production_, well, then this one _can go to Mars_.

> [!IMPORTANT]
> [**libboostasio_2025-12-23.zip**](https://github.com/user-attachments/files/24314828/libboostasio_2025-12-23.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -925,7 +925,7 @@ channel2.declareExchange("my-exchange");
 
 Now, if an error occurs with declaring the queue, it will not have consequences
 for the other call. But this comes at a small price: setting up the extra channel
-requires and extra instruction to be sent to the RabbitMQ server, so some extra
+requires an extra instruction to be sent to the RabbitMQ server, so some extra
 bytes are sent over the network, and some additional resources in both the client
 application and the RabbitMQ server are used (although this is all very limited).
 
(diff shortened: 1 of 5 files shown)


paolopas:
#547 fixed

